PR #11, opened 2023-12-23
showmore is fixed in leanblueprint

utensil:
More context: https://leanprover.zulipchat.com/#narrow/stream/412902-Polynomial-Freiman-Ruzsa-conjecture/topic/.22unfolding.20triangles.22.20in.20blueprint


kbuzzard:
OK let's go for it :-)

PR #27, opened 2024-04-30
fix typo, add to .gitignore

digama0:
* `$f K\to L$ -> $f:K\to L$`
* Added some more latex auxiliary files to the gitignore
* Is `GlobalLanglandsConjectures/notes.pdf` supposed to be checked in? I get local modifications on that file because it's checked in even though it's in the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,4 +15,7 @@ blueprint/lean_decls
 *.log
 *.out
 *.pdf
+*.fls
+*.fdb_latexmk
+*.synctex.gz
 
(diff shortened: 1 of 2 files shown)


kbuzzard:
Oh yeah those notes were for my students! Thanks.

kbuzzard:
What thing are you using which is causing trailing spaces to be deleted? Clearly I should be using it too.

digama0:
It's a vscode setting, you can enable it in `.vscode/settings.json`, [as mathlib does](https://github.com/leanprover-community/mathlib4/blob/master/.vscode/settings.json#L9). (You might also want to consider the other settings from that file.)

kbuzzard:
Thanks! I've just added one here too.

PR #28, opened 2024-04-30
feat: discriminant of P.FreyCurve

### LINKED ISSUE #21: FLT.FreyCurve
In `Basic/Reductions` the definition of the Frey curve is incomplete; we need to check that the discriminant is nonzero, and I've made the global change of variables to ensure that the curve is minimal at 2.

Ruben-VandeVelde:
Fixes #21.

(I think)


kbuzzard:
Thanks!

PR #30, opened 2024-05-01
feat: define the Frey curve

trivial1711:
Fill in the `sorry`s in the definition of the Frey curve in `FLT.Basic.Reductions`


trivial1711:
I didn't realize Ruben VandeVelde already did this

PR #32, opened 2024-05-01
define "locally free" and state that it is equivalent to "flat" for finitely presented modules

trivial1711:


kbuzzard:
@trivial1711 Are you done with this or is this WIP?

trivial1711:
I was going to do this by incrementally setting up pull requests to this repository (like this one) that have some `sorry`s. But after thinking about it a bit more, I think I might close this and try to PR the necessary proofs to mathlib directly. What do you think?

trivial1711:
(The reason is that a lot of the necessary ingredients to the proof need to go into the respective parts of mathlib)

PR #40, opened 2024-05-02
Homogenise and tidy .bib file

pitmonticone:


kbuzzard:
I get my .bib info from MathSciNet, which produces it the way it currently is. If I merge this then in a few months time it'll just be a mix of styles. Furthermore I've got to manually check that you didn't mangle any current references. So I'm reluctant to merge this.

pitmonticone:
I've done it automatically using [this tool](https://github.com/FlamingTempura/bibtex-tidy) without changing anything in the content apart from sorting of fields. 

I can easily run it periodically if you want or maybe do it once and for all at the end.

kbuzzard:
I think it's probably easier to do it once and for all at the end.

PR #44, opened 2024-05-02
Substitute `\cite` to `\ref` in `chtopbestiary.tex`

pitmonticone:


kbuzzard:
I guess we have both Mazur as a reference and a citation! I will need to think about what we actually want to say here.

pitmonticone:
I see. I've opened it because there is no bib entry labelled as `mazur`, only a theorem labelled as such.

kbuzzard:
Thanks. In the end I've rewritten the section and added both the citation and the reference in 6d3b360

Issue #37, opened 2024-05-01
Package `fontspec` Error: font "Latin Modern Math" cannot be found


pitmonticone:
I've tried to locally compile `print.tex` and it returns the following error: 

> ! Package fontspec Error: The font "Latin Modern Math" cannot be found.

In order to take a look at the `print.pdf` I had to temporarily comment out the font specifications in `print.tex` as follows: 

```latex
% [...]

\usepackage[warnings-off={mathtools-colon,mathtools-overbracket}]{unicode-math}
\usepackage{fontspec}
% \setmathfont{Latin Modern Math}
% \setmathfont[range=\varnothing]{Asana-Math.otf}
% \setmathfont[range=\pitchfork]{Asana-Math.otf}
% \setmathfont[range=\intprod]{Asana-Math.otf}
% \setmathfont[range=\int]{Latin Modern Math}

% [...]
```

I've already checked if the font package is properly installed with 

```console
tlmgr info lm-math
```

and, as expected, it returned the usual: 

```nothing
package:     lm-math
category:    Package
shortdesc:   OpenType maths fonts for Latin Modern
longdesc:    Latin Modern Math is a maths companion for the Latin Modern family of fonts, in OpenType format. For use with LuaLaTeX or XeLaTeX, support is available from the unicode-math package.
installed:   Yes
revision:    67718
sizes:       doc: 149k, run: 721k
relocatable: No
cat-version: 1.959
cat-license: gfl
cat-topics:  font font-maths font-otf
cat-contact-home: https://www.gust.org.pl/projects/e-foundry/latin-modern
collection:  collection-fontsrecommended
```

pitmonticone:
Fixed substituting with `\setmainfont{latinmodern-math.otf}`. 

Thank you @utensil

PR #45, opened 2024-05-02
Fix macOS (?) issue with latin modern math font

pitmonticone:


kbuzzard:
Can you explain what's going on here? I guess I should try seeing if this PR compiles locally for me. What changes would you expect to happen if I merge it?

pitmonticone:
Sure. Everything is explained in this (now closed) issue #37

kbuzzard:
OK, leanblueprint still seems to be working for me after this change, so I'll take your word for it that it's better for mac users.

PR #50, opened 2024-05-03
half of MatrixRing.isCentralSimple

Ruben-VandeVelde:
Not sure when/if I'll get around to the other half.


FLT/for_mathlib/IsCentralSimple.lean
```diff
@@ -30,8 +30,26 @@ structure IsCentralSimple
...
+    split_ifs with h
+    · subst h
+      specialize hd (stdBasisMatrix k i (1 : K))
+      rw [← ext_iff] at hd
+      specialize hd k i
+      simpa using hd
+    · specialize hd (stdBasisMatrix j i (1 : K))
+      rw [← ext_iff] at hd
+      specialize hd j j
+      simpa [Ne.symm h] using hd
```
eric-wieser, line 51:
```suggestion
  is_central d hd := by
    rw [Subring.mem_center_iff] at hd
    convert mem_range_scalar_of_commute_stdBasisMatrix (M := d) fun i j hij => hd _
    simp_rw [Set.mem_range, eq_comm]
    rfl
```
Change the definition to use `Set.range` instead and you can knock three more lines off the proof.

Ruben-VandeVelde: Thanks!

kbuzzard:
Thanks to both of you!

PR #29, opened 2024-04-30
feat: fill in of_not_FermatLastTheorem_coprime_p_ge_5

Ruben-VandeVelde:

PR #51, opened 2024-05-03
feat: simplify proof of ¬FreyPackage -> FLT

digama0:
* [ ] Depends on: #29

Simplifies the reduction proof. As a bonus, the proof is now even closer to the text proof, proceeding by a sequence of reductions rather than constructing a function and doing case splits in all the theorems.


kbuzzard:
Thanks a lot Mario!

PR #56, opened 2024-05-05
fix statement of ZHat.multiples

dagurtomas:
The old statement didn't make sense, also fixes a typo in the blueprint.


kbuzzard:
Thanks!

PR #57, opened 2024-05-06
Update to the newer lakefile format

eric-wieser:
The `autoImplicit` setting was not actually being applied, as the lines were commented out.

Uncommenting the lines would have resulted in breaking Std files in VSCode.
`leanOptions` is now (4.6.0?) the preferred spelling of `moreServerArgs`, which does not have this issue.

`autoImplicit` flags that one statement appears to be garbage.

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -1,27 +1,12 @@
 import Lake
 open Lake DSL
 
-def moreServerArgs := #[
-  "-Dpp.unicode.fun=true", -- pretty-prints `fun a ↦ b`
-  "-Dpp.proofs.withType=false", -- no idea what this does
-  "-DautoImplicit=false", -- attempt to switch off auto-implicit
-  "-DrelaxedAutoImplicit=false" -- attempt to switch off relaxed auto-implicit
-]
-
--- These settings only apply during `lake build`, but not in VSCode editor.
-def moreLeanArgs := moreServerArgs
-
--- These are additional settings which do not affect the lake hash,
--- so they can be enabled in CI and disabled locally or vice versa.
--- Warning: Do not put any options here that actually change the olean files,
--- or inconsistent behavior may result
-def weakLeanArgs : Array String :=
-  if get_config? CI |>.isSome then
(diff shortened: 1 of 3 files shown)


lakefile.lean
```diff
@@ -36,6 +21,4 @@ require «doc-gen4» from git
 lean_lib FLT where
   globs := #[
     .andSubmodules `FLT
-    ]
---  moreLeanArgs := moreLeanArgs
---  weakLeanArgs := weakLeanArgs
```
eric-wieser, line 41: Previously these weren't actually taking effect.

kbuzzard: Oh that would explain why autoimplicits were on anyway :-) Thanks for fixing things up!

kbuzzard:
I tried `leanblueprint all` on this branch and got

```
error: ././.lake/packages/checkdecls/././Main.lean:13:36-13:48: too many explicit universe levels for 'Lake.LakeOptions.mkLoadConfig'
error: ././.lake/packages/checkdecls/././Main.lean:14:11-14:73: failed to synthesize instance
  ToString Empty
error: ././.lake/packages/checkdecls/././Main.lean:15:17-15:89: invalid field 'root', the environment does not contain 'Lake.EResult.root'
  ws
has type
  EResult Nat Log Workspace
error: ././.lake/packages/checkdecls/././Main.lean:14:60-14:72: invalid dotted identifier notation, unknown identifier `Lake.Log.eio` from expected type
  Log
error: Lean exited with code 1
Some build steps logged failures:
- Building Main
error: build failed
Command 'lake exe checkdecls blueprint/lean_decls' returned non-zero exit status 1.
```

kbuzzard:
I needed a heartbeat bump but everything else seems to work, and the heartbeat bump is in API which is being rewritten anyway. Thanks!

PatrickMassot:
This isn’t the new lakefile format! Where is the toml version?

eric-wieser:
@PatrickMassot: this PR was made prior to the bump to lean 4.8.0

PR #58, opened 2024-05-06
Adjust namespaces for `Coalgebra`, and remove a heartbeat bump

eric-wieser:
Writing a `CoalgebraStruct` first allows a few helper lemmas to be written before the main instance.

This also removes the `Coalgebra` and `Bialgebra` components from the namespaces.


kbuzzard:
Thanks!

PR #68, opened 2024-05-07
Fix `hyperref` warning

pitmonticone:
Fix `hyperref` warnings such as:

```console
Package hyperref Warning: Token not allowed in a PDF string (Unicode):
(hyperref)                removing `math shift' on input line 419.
```

This warning is generated `hyperref` package when it encounters a math symbol in a section heading or caption, which is not allowed in a PDF string. It automatically removes the math symbol, but it still generates a warning.

Using `\texorpdfstring{$X$}{X}` prevents the warning.


kbuzzard:
Oh nice! Thanks!

PR #73, opened 2024-05-09
lowestTerms

Ruben-VandeVelde:


kbuzzard:
Oh wow thanks! Some of this stuff can be useful on the Galois side of the example too.

AbdulShabazz:
@kbuzzard Although this is perhaps a trivial instance, I would strongly advise against dropping unused lemma and definitions et al., if they pertain to FLT, and or may later be re-derived, or can be properly categorized and reused elsewhere in Mathlib!

Ruben-VandeVelde:
@AbdulShabazz what?

FLT/AutomorphicRepresentation/Example.lean
```diff
@@ -338,6 +357,153 @@ noncomputable abbrev i₁ : ℚ →ₐ[ℤ] QHat := Algebra.TensorProduct.includ
 lemma injective_rat :
     Function.Injective i₁ := RingHom.injective i₁.toRingHom
 
+theorem PNat.lcm_comm (m n : ℕ+) : PNat.lcm m n = PNat.lcm n m := PNat.eq <| by
+  simp [Nat.lcm_comm]
+
+lemma lowestTerms (x : QHat) : (∃ N z, IsCoprime N z ∧ x = (1 / N : ℚ) ⊗ₜ z) ∧
+    (∀ N₁ N₂ z₁ z₂,
+    IsCoprime N₁ z₁ ∧ IsCoprime N₂ z₂ ∧ (1 / N₁ : ℚ) ⊗ₜ z₁ = (1 / N₂ : ℚ) ⊗ₜ[ℤ] z₂ →
+      N₁ = N₂ ∧ z₁ = z₂) := by
```
kbuzzard, line 366: This is surely two different lemmas? I don't know why I made it just one. The second clause doesn't even mention x. I guess it could be an `∃!` but in practice is this even more difficult to use?

PR #74, opened 2024-05-10
Multiplication for Hurwitz numbers

Ruben-VandeVelde:
Carefully worked out on paper, so might well be wrong :)


kbuzzard:
I'm prepared to take the gamble :-) Fix later if necessary -- that's the point of computers.

PR #78, opened 2024-05-13
Add missing `leanok` to lemma 5.13

Rida-Hamadani:
[Lemma 5.13](https://imperialcollegelondon.github.io/FLT/blueprint/sect0004.html#QHat.canonicalForm) is done [here](https://github.com/ImperialCollegeLondon/FLT/blob/1ae6d6c0b3895b2260852cb93a7ac150f3ab9138/FLT/AutomorphicRepresentation/Example.lean#L270-L294) but the check mark is not showing.


kbuzzard:
Looks like we put leanok on the proof but not the statement. Thanks!

PR #80, opened 2024-05-14
Norm

Ruben-VandeVelde:
Depends on #79


kbuzzard:
Thanks for proving the theorems so now we know the definitions are good :-)

PR #81, opened 2024-05-15
Add missing `leanok` to lemma 6.2

Rida-Hamadani:
[Lemma 6.2](https://imperialcollegelondon.github.io/FLT/blueprint/sect0005.html#MatrixRing.isCentralSimple) is done [here](https://github.com/ImperialCollegeLondon/FLT/blob/3acd39171d3c6b60a09bf7730eb8d08a5bbf120b/FLT/for_mathlib/IsCentralSimple.lean#L52-L102) but the check mark is missing.  This is another case where the `\leanok` was added to the proof but not to the statement.


Ruben-VandeVelde:
Fwiw, I also covered this in #83 before I noticed this PR

Rida-Hamadani:
@Ruben-VandeVelde Thanks for letting me know, I will close this to avoid merge conflicts.

PR #84, opened 2024-05-16
Crazy affine

Whysoserioushah:


kbuzzard:
This PR is thousands of lines long; it is not possible for me to review that amount of code in one go.

But actually I'm going to close this PR because I don't think that the FLT repo is a good place to have the theory of Hopf algebras: they really should be in mathlib. Other people at Imperial are moving this material over to mathlib but this PR is still very useful because it shows us the API which we need.

PR #92, opened 2024-05-30
Prove Hurwitz.exists_near

Ruben-VandeVelde:
Also avoid using two variables called 'z'.


kbuzzard:
Oh nice! I should probably solve the topology problem for adeles because we're getting close to being able to give an example of a space of quaternionic modular forms, except that I can't make the definition yet...

PR #97, opened 2024-06-11
Completed a few sorries in HIMExperiments/module_topology.lean

scholzhannah:
The following proofs in the file HIMExperiments/module_topology.lean are done: 
- Module.homeomorphism_equiv (some trivial sorries)
- Module.topology_self (this was added to prove the next lemma)
- Module.prod_canonical (one of two identical cases is not filled in yet, which should probably be done after cleanup of the proof)

### DIFF
--- a/FLT/HIMExperiments/module_topology.lean
+++ b/FLT/HIMExperiments/module_topology.lean
@@ -1,6 +1,5 @@
 import Mathlib.RingTheory.TensorProduct.Basic -- we need tensor products of rings at some point
 import Mathlib.Topology.Algebra.Module.Basic -- and we need topological rings and modules
-
 /-
 
 # The "module topology" for a module over a topological ring.
@@ -51,7 +50,7 @@ and target.
 --   refine Continuous.mul (Continuous.comp (continuous_apply _) (continuous_fst)) ?_
 --   exact (Continuous.comp (continuous_apply _) (continuous_snd))
 
-variable (A : Type*) [CommRing A] [TopologicalSpace A] [TopologicalRing A]
+variable (A : Type*) [CommRing A] [hA : TopologicalSpace A] [TopologicalRing A]
 
 -- let M be an A-module
 variable {M : Type*} [AddCommGroup M] [Module A M]
@@ -107,10 +106,10 @@ def Module.homeomorphism_equiv (e : M ≃ₗ[A] N) :
   -- all the sorries should be formal.
   { toFun := e
     invFun := e.symm
-    left_inv := sorry
-    right_inv := sorry
-    continuous_toFun := sorry
-    continuous_invFun := sorry
+    left_inv := e.left_inv
+    right_inv := e.right_inv
+    continuous_toFun := Module.continuous_linear A e
+    continuous_invFun := Module.continuous_linear A e.symm
   }
 
 -- sanity check: the topology on A^n is the product topology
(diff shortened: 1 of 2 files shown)


FLT/HIMExperiments/module_topology.lean
```diff
@@ -107,10 +106,10 @@ def Module.homeomorphism_equiv (e : M ≃ₗ[A] N) :
...
   { toFun := e
     invFun := e.symm
-    left_inv := sorry
-    right_inv := sorry
-    continuous_toFun := sorry
-    continuous_invFun := sorry
+    left_inv := e.left_inv
+    right_inv := e.right_inv
+    continuous_toFun := Module.continuous_linear A e
+    continuous_invFun := Module.continuous_linear A e.symm
```
kbuzzard, line 112: Very nice!

FLT/HIMExperiments/module_topology.lean
```diff
@@ -171,7 +238,7 @@ noncomputable def LinearMap.add (M : Type*) [AddCommGroup M] [Module A M] :
 
 -- Now say we have a non-commutative `A`-algebra `D` which is free of finite type.
 
-variable (D : Type*) [Ring D] [Algebra A D] [Module.Finite A D] [Module.Free A D]
+variable (D : Type*) [Ring D] [Algebra A D]
```
kbuzzard, line 241: I am more nervous about the bilinear map sorry without these instances -- can you put them back, but just above ` Module.continuous_bilinear` ?

FLT/HIMExperiments/module_topology.lean
```diff
@@ -137,13 +136,81 @@ example (ι : Type*) [Finite ι] :
...
     @instTopologicalSpaceProd M N (Module.topology A) (Module.topology A) =
     (Module.topology A : TopologicalSpace (M × N)) := by
   -- the goal is to prove that an iInf equals an inf (of two topologies).
-  unfold instTopologicalSpaceProd -- you can probably delete this later
-  sorry
+  apply le_antisymm
+  · apply le_iInf
+    intro f
+    rw [← continuous_iff_le_induced]
+    have : ⇑f = fun ⟨m, n⟩ ↦ f (⟨m, 0⟩) + f (⟨0, n⟩) := by
```
kbuzzard, line 167: So far so good. In my `example` I now applied a lemma from the library, and you have decided to prove the analogous lemma *in the middle of another proof*. This is not normally a good strategy -- those lemmas should be pulled out. You should formalise the statement of what you need, then check with loogle and `exact?` to see if it's there, and if it's not there you should think of a good name (e.g. `LinearMap.continuous_on_prod`) and then write it separately outside this proof. Lots of short proofs is always better than one long one.

FLT/HIMExperiments/module_topology.lean
```diff
@@ -137,13 +136,81 @@ example (ι : Type*) [Finite ι] :
...
+      := by sorry -- this will be exactly the same as above so better do the cleanup first
+    apply @Continuous.add _ _
+      (@instTopologicalSpaceProd M N (Module.topology A) (Module.topology A)) _ _ _ _ _ h1 h2
+  · apply le_inf
+    · rw [induced_iInf]
+      apply le_iInf
+      intro f
+      rw [induced_compose]
+      exact iInf_le (fun (f :  M × N →ₗ[A] A) ↦ TopologicalSpace.induced (⇑f) inferInstance)
+        (LinearMap.lcomp A A (LinearMap.fst A M N) f)
```
kbuzzard, line 207:
```suggestion
      exact iInf_le _ (LinearMap.lcomp _ _ (LinearMap.fst _ _ _) _)
```
I can imagine that you may well have filled in a lot of those holes when you were golfing this proof, but I'm just observing that if you want you can drop many of them, because `exact` is good at unification.

kbuzzard (commented):
Thanks for your PR! I left a review. Any questions you can either ask here or we can chat tomorrow. If you resolve a comment (e.g. you adopt a suggestion of mine) then you can mark the conversation as resolved. When you're ready for a second round change the tag to `awaiting-review`.

FLT/HIMExperiments/module_topology.lean
```diff
@@ -137,13 +136,81 @@ example (ι : Type*) [Finite ι] :
...
+    have hi : i = LinearMap.lsmul A A (i 1) := by
+      ext
+      simp only [LinearMap.lsmul_apply, smul_eq_mul, mul_one]
+    rw [hi]
+    apply continuous_const.mul continuous_id
+    -- a linear map should always be continuous in this setting right?
+  · apply sInf_le
+    rw [Set.mem_range]
+    use LinearMap.id
+    exact induced_id
```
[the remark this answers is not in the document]
scholzhannah: Wow that's super short. I didn't know you could put a `show` into a `rw` like that. Thanks!

FLT/HIMExperiments/module_topology.lean
scholzhannah: We'll try to figure it out with these conditions then.

kbuzzard (approved):
Thanks a lot!

PR #100, opened 2024-06-12
finite codimension

digama0:
Working towards sorry-free definition of automorphic form for GL_n/Q. Work carried out in the agreeable environment of the HIM in Bonn.


kbuzzard:
Thanks a lot Mario!

PR #103, opened 2024-06-14
GLnDefs some sorrys filled in

TBUGTB:
Some progress with the sorrys in GLnDefs. 

There is still one bit missing in the definition of `lift'`. I tried stating a compatibility statement for e.symm and the Lie bracket, but somehow Lean has difficulties finding the right HMul instance. To be followed up...


kbuzzard:
Thanks!

PR #110, opened 2024-06-21
Hecke rings

CBirkbeck:
Fist go at Hecke rings. Still lots of clean up to do and lots of sorrys left. Associativity is tricky, that said all the sorrys were known in the 1980s.....


CBirkbeck:
Yeah I think what I want for modular forms is not what you want here.  So yes feel free to close it.

CBirkbeck:
But can you not close it until Monday, so I can double check I have it saved? Or that I can save the code? (I expect it's a non issue, i just don't remember how it works)

PR #112, opened 2024-07-06
Blueprint: Fix label error

pitmonticone:
Fix LaTeX error: 

> ERROR: Label 'AutomorphicForm.GLn.weight' could not be resolved


kbuzzard:
Thanks!

PR #114, opened 2024-07-26
Bump GitHub actions

pitmonticone:
## Changes

- [x] Bump actions/cache from 3 to 4
- [x] Bump actions/upload-pages-artifact from 1 to 3
- [x] Bump actions/deploy-pages from 1 to 4
- [x] Bump actions/checkout from 2 to 4

## Effects

- Solved all warnings 
- Disk space usage seems to be reduced significantly, but I think it's nothing but a change in the variable displayed (from uncompressed to compressed size)

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -29,7 +29,7 @@ jobs:
     name: Build project
     steps:
       - name: Checkout project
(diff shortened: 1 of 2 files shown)


kbuzzard (commented):
Thanks!

PR #117, opened 2024-08-04
Deduplicate upload action

YaelDillies:
There is no reason for "Upload docs & blueprint artifact" action to exist twice, and indeed its duplication is currently causing CI to fail.


kbuzzard:
Let's give it a try!

PR #120, opened 2024-09-04
Bump lean and mathlib

pitmonticone:
- [x] Bump dependencies
- [x] Upgrade deprecated lemmas
- [x] `sorry` a few problematic proof steps
- [x] Add `#exit` in Frobenius as prescribed


kbuzzard:
Thanks a lot! Some proofs are now broken but my instinct here is to merge first and ask questions later.

PR #138, opened 2024-09-28
Add update workflow

pitmonticone:
In brief, this workflow runs every 3 days at 8 AM UTC or can be triggered manually. It checks out the code, updates Lean project dependencies, and handles the outcome. If the update is successful, it creates a pull request (not direct push yet for precautionary reasons); if it fails, it opens an issue to notify us.

Of course let me know if you want to change anything about the scheduling.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,0 +1,22 @@
+name: Update Dependencies
+
+on:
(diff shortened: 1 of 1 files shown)


kbuzzard:
Thanks!

PR #158, opened 2024-10-03
Improve the blueprint graph a bit

Ruben-VandeVelde:


kbuzzard:
Thanks a lot! Somehow I still haven't figured out the right workflow for all of this. I write the LaTeX, I write the Lean statements, I glue the Lean to the LaTeX, I wish it were a little easier, I sometimes only do half a job. Thanks a lot for doing the other half.

PR #190, opened 2024-11-01
Fix failed bump

### LINKED ISSUE #189: Updates available but manual intervention required
Try `lake update` and then investigate why this update causes the lean build to fail.
Files changed in update:

pitmonticone:
null

Closes #189


kbuzzard:
Thanks a lot for debugging these!

PR #194, opened 2024-11-04
fill in `valuation_comap`

erdOne:


erdOne:
The right thing to do is probably to generalize `HeightOneSpectrum.valuation` to take in any ideal...

kbuzzard:
So what is the 6-adic valuation of 2, of 3 and of 6?

kbuzzard:
Thanks!

PR #211, opened 2024-11-08
Proved B is finite

### LINKED ISSUE #202: The integral closure of a Dedekind domain in a finite separable extension is finite
Is it true that if A is a Dedekind domain with field of fractions K, and if L/K is a finite separable extension, then the integral closure B of A in L is a finite A-module? I'm assuming so but was reluctant to add it as an assumption if it was not true in this generality. This is

```lean
example : Module.Finite A B := by sorry -- I assume this is correct
```

in `FLT.DedekindDomain.FiniteAdeleRing.BaseChange` .

4hma4d:
Proved `Module.Finite A B` in `BaseChange.lean`
Closes #202

### DIFF
--- a/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
+++ b/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
@@ -36,9 +36,21 @@ variable [IsDomain B]
 -- example : Algebra.IsIntegral A B := IsIntegralClosure.isIntegral_algebra A L
 variable [Algebra.IsIntegral A B]
 
--- I can't find in mathlib the assertion that B is a finite A-moduie.
--- It should follow from L/K finite.
-example : Module.Finite A B := by sorry -- I assume this is correct
+example: Module.Finite A B := by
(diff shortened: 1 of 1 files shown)


kbuzzard:
Thanks a lot! This is great. I gave you a code review because I'm not sure I've seen you around before, but the code is fine; I was just showing you some tricks. You might want to consider upstreaming this result to mathlib -- I couldn't find it but I'm pretty sure that the experts there will know if it's there, and if it's not then I'm pretty sure they'll want it.

kbuzzard (approved):

4hma4d:
Thanks for the review! I looked through mathlib again, and I found that B is noetherian is already there, so the proof becomes trivial 😅

4hma4d:
I submitted https://github.com/leanprover-community/mathlib4/pull/18842

kbuzzard:
Thanks! Probably we should wait for that to get merged and bump mathlib and modify this PR to use it.

kbuzzard:
Actually I just copied your mathlib PR into this repo :-) Many thanks for sorting this out!

PR #223, opened 2024-11-18
feat: compute the distributive Haar character of `ℝ`, `ℂ`, `ℤ_[p]` and `ℚ_[p]`

YaelDillies:
Co-authored-by: Javier López-Contreras <javierlc2000@gmail.com>

---

- [x] depends on: #263


FLT/HaarMeasure/ModularCharacter.lean
```diff
@@ -4,6 +4,23 @@ Released under Apache 2.0 license as described in the file LICENSE.
...
+
+/-!
+# The modular character of Haar measures
+
+Given a group `G` acting on an measurable additive commutative group `A`, and an element `g : G`,
+one can pull back the Haar measure `μ` of `A` along the map `(g • ·) : A → A` to get another Haar
+measure `μ'` on `A`. By unicity of Haar measures, there exists some nonnegative real number `r` such
+that `μ' = r • μ`. We can thus define a map `modularHaarChar : G → ℝ≥0` sending `g` to its
+associated real number `r`. Furthermore, this number doesn't depend on the Haar measure `μ` we
+started with, and `modularHaarChar` is a group homomorphism.
```
kbuzzard, line 18: In the literature "modular character" means something else unfortunately, so I don't think we should call it this. In the literature, `A = Additive G` and the action is just translation, which is not an action of G on the group A; it is a left action on the set A which commutes with the right action of A on itself by addition. I think we need to stick to the literature when it comes to naming. I don't know what the name of what you're defining is but you are assuming `DistribMulAction` so I don't think it can be "modular character". See for example https://math.stackexchange.com/questions/2237765/how-is-the-modular-character-usually-defined for the sign confusion in the literature, but you're doing something else as far as I can see.

FLT/ForMathlib/MeasurableSpacePadics.lean
```diff
@@ -0,0 +1,6 @@
+import Mathlib
+
+variable {p : ℕ} [Fact p.Prime]
+
+instance : MeasurableSpace ℚ_[p] := sorry
+instance : BorelSpace ℚ_[p] := sorry
```
kbuzzard, line 6: I think that this might be over-reach for this PR. Can we just get the character I need in?

FLT/HaarMeasure/Map.lean
```diff
@@ -14,22 +17,42 @@ is multiplicative on the group of topological group endomorphisms of `G`.
...
+noncomputable
 def TopologicalRing.AddHaarMeasure.scale (R : Type*) [Ring R] [TopologicalSpace R]
-    [TopologicalRing R] [LocallyCompactSpace R] :
-    Rˣ →* ℝ := by
+    [TopologicalRing R] [MeasurableSpace R] [BorelSpace R] [LocallyCompactSpace R] :
+    Rˣ →* ℝ≥0 := modularHaarChar R
+
+open TopologicalRing AddHaarMeasure
+
+lemma scale_complex (z : ℂˣ) : scale ℂ z = ‖(z : ℂ)‖₊ ^ (-2 : ℤ) := by
```
kbuzzard, line 37: I think that this (reals/complexes) is out of scope for this PR.

FLT/Mathlib/Data/ENNReal/Inv.lean
```diff
@@ -0,0 +1,8 @@
+import Mathlib.Data.ENNReal.Inv
+
+namespace ENNReal
+
+protected lemma mul_div_cancel_right {b : ℝ≥0∞} (hb₀ : b ≠ 0) (hb : b ≠ ∞) (a : ℝ≥0∞) :
+    a * b / b = a := by rw [ENNReal.mul_div_right_comm, ENNReal.div_mul_cancel hb₀ hb]
```
kbuzzard, line 6: Can you clarify whether (a) this is already a PR to mathlib and (b) whether you are going to make this PR or are leaving it so someone else? Same comment for every file in `FLT/Mathlib`.

kbuzzard (commented):
Here are two ideas which are (to me) distinct.

1) Given a multiplicative group G acting as a DistribMulAction on an additive group A (e.g. A is a ring and G=A^*) with an additive Haar measure, there's a character G -> R_{>0}. Note that the key axiom is `g . (a1 + a1) = g . a1 + g . 2`, and G acts on A via additive group isos. Everything can be commutative here and the answer can still be interesting.

2) Given a multiplicative group G acting on a multiplicative group H (e.g. H=G)  with [IsScalarTower G H H] i.e. the axiom `(g • h1) * h2 = g • (h1 * h2)`, and a multiplicative Haar measure on H, there's a character G -> R_{>0}. Note that G acts on H via things which aren't group isos, but they are bijections. Note also that nontriviality of this character implies that left and right Haar measures on H are different and in particular that H is not commutative.

"Modular character" is reserved in the literature for (2). I want (1). In this PR right now we have two files `FLT/ForMathlib/DomMulActMeasure.lean` and `FLT/HaarMeasure/ModularCharacter.lean` both with docstrings which seem to say basically the same thing (and both of which seem to be (1)), and both of which seem to use the words "modular character" for (1). We need to call it something else. How about `distribHaarCharacter`?

FLT/Mathlib/Data/ENNReal/Inv.lean
YaelDillies: My plan is to equip FLT with [the upstreaming dashboard which I have developed for LeanCamCombi](https://yaeldillies.github.io/LeanCamCombi/upstreaming) (warning: not looking at its best right now, Javier and I will fix it later today), meaning that anyone will be able to upstream those files to mathlib asynchronously.

I personally haven't planned yet to PR that to mathlib, but once there are enough things in the Mathlib folder I will open series of PRs with that material.

FLT/ForMathlib/MeasurableSpacePadics.lean
YaelDillies: What is the character you need? Two weeks you said we should compute this canonical norm for the reals, complexes, p-adic numbers and integers, which is what Javier and I tried doing

YaelDillies:
> Here are two ideas which are (to me) distinct. [...]

Yes, Javier and I ended up understanding what you meant by "This is not what the literature calls modular characters". Will update this PR later today

FLT/ForMathlib/MeasurableSpacePadics.lean
kbuzzard: I think this PR should just stick to getting `DistribHaarCharacter` over the line. As you say, I need the p-adics, reals and complexes, but those can come later.

FLT/ForMathlib/MeasurableSpacePadics.lean
```diff
@@ -0,0 +1,50 @@
...
+        IsUltrametricDist.isOpen_closedBall (0 : ℚ_[p]) one_ne_zero
+
+noncomputable instance : MeasureSpace ℚ_[p] := ⟨addHaarMeasure Padic.unitBall_positiveCompact⟩
+
+instance : IsAddHaarMeasure (volume : Measure ℚ_[p]) := isAddHaarMeasure_addHaarMeasure _
+
+lemma Padic.volume_closedBall : volume {x : ℚ_[p] | ‖x‖ ≤ 1} = 1 := addHaarMeasure_self
+
+lemma PadicInt.volume_addSubgroupClosure (x : ℤ_[p]ˣ) :
+    volume (AddSubgroup.closure {x.1} : Set ℤ_[p]) = ‖x.1‖₊⁻¹ := sorry
```
kbuzzard, line 47: I think this sorry is false. Firstly, if x is a unit in Z_p then multiplication by x is a bijection Z_p -> Z_p so whatever kind of norm you are putting on x (the p-adic norm, or the one coming from DistribHaarChar etc), the answer will be 1. Conversely, `AddSubgroup.closure {x.1}` will be a countable set so whatever volume you are putting on it, the answer will be 0.

FLT/ForMathlib/MeasurableSpacePadics.lean
```diff
@@ -0,0 +1,50 @@
+import Mathlib.MeasureTheory.Measure.Haar.Basic
```
kbuzzard, line 1: I was planning on moving away from `ForMathlib` and adopting Yael's `Mathlib` approach. Can we move this into the `Mathlib` directory?

kbuzzard (commented):
Thanks! This mostly looks fine, but I have an issue with one of the sorrys.

FLT/ForMathlib/MeasurableSpacePadics.lean
YaelDillies: Ah I think the issue is that we thought `a • top = closure {a}`, but you are right that the second one only contains very specific sequences of digits, namely the multiples of `a`, while the first one contains many more things.

We can just use `a • top` everywhere

kbuzzard: This will solve the mathematical problem, but the other thing I'm saying is that `a • top = top` because if u is a unit in a ring R then uR=R because if vu=1 then R = vuR subseteq uR subseteq R

javierlcontreras: Yep, this is my fault. I wanted `u \in nonZeroDivisors \Z_[p]`

FLT/ForMathlib/MeasurableSpacePadics.lean
javierlcontreras: Yep!

FLT/HaarMeasure/ConcreteCharCalculations.lean
```diff
@@ -0,0 +1,79 @@
...
+import FLT.Mathlib.LinearAlgebra.Determinant
+import FLT.Mathlib.MeasureTheory.Group.Action
+import FLT.Mathlib.NumberTheory.Padics.PadicIntegers
+import FLT.Mathlib.NumberTheory.Padics.MeasurableSpacePadics
+import FLT.Mathlib.RingTheory.Norm.Defs
+
+open Real Complex Padic MeasureTheory Measure NNReal Set TopologicalSpace
+open scoped Pointwise ENNReal Localization
+
+lemma distribHaarChar_complex (z : ℂˣ) : distribHaarChar ℂ z = ‖(z : ℂ)‖₊ ^ (-2 : ℤ) := by
```
kbuzzard, line 18: Oh crap, really?? The definition of `distribHaarChar` is out by a ^{-1} then :-(

FLT/HaarMeasure/ConcreteCharCalculations.lean
```diff
@@ -0,0 +1,79 @@
...
+    simp [Algebra.norm_complex_apply, normSq_eq_norm_sq, zpow_ofNat]
+  have := addHaar_preimage_linearMap (E := ℂ) volume
+    (f := (LinearMap.mul ℂ ℂ z).restrictScalars ℝ)
+  simp [key] at this
+  convert this _ using 2
+  · symm
+    exact Set.preimage_smul z _
+  · simp [← Real.ennnorm_eq_ofReal_abs, ← Complex.norm_eq_abs, zpow_ofNat]
+
+lemma distribHaarChar_real (x : ℝˣ) : distribHaarChar ℝ x = ‖(x : ℝ)‖₊⁻¹ := by
```
kbuzzard, line 34: The whole point is that this `distribHaarChar` is a way of recovering the norm, not a way of recovering the inverse of the norm.

YaelDillies: I see a few options then:
* Use `(distribHaarChar G x)⁻¹`. Likely painful
* Keep assuming that `G` acts on `A` but make `distribHaarChar G` a group hom with domain `Gᵐᵒᵖ`. Likely painful
* Keep `distribHaarChar G` a group hom with domain `G` but assume that `Gᵐᵒᵖ` acts on `A` instead of `G`. Probably the best solution?

Anyway, I think this isn't a discussion for the current PR. The definition is already on master, this PR only computes it, and the refactor to invert everything should be its own PR and can be done after the current PR.

kbuzzard: I forced it to be an issue in this PR by fixing the definition in #274 (now merged).

FLT/HaarMeasure/ConcreteCharCalculations.lean
```diff
@@ -0,0 +1,83 @@
...
+    (by simp [Z, closedBall_openAddSubgroup, Metric.closedBall, Padic.volume_closedBall])
+    (by simp [Z, closedBall_openAddSubgroup, Metric.closedBall, Padic.volume_closedBall]) ?_
+  let H := (x : ℚ_[p]) • Z
+  rw [Units.smul_def]
+  change volume (H : Set ℚ_[p]) = _
+  have : (Z.addSubgroupOf H).FiniteIndex := finiteIndex_smul _
+  rw [← index_mul_addHaar_addSubgroup_eq_addHaar_addSubgroup (μ := volume) (H := Z)]
+  · congr 1
+    exact mod_cast relindex_smul x
+  · sorry
```
kbuzzard, line 66: This `sorry` seems to be the false claim that the closed ball centre 0 radius 1 is a subset of the closed ball centre 0 radius 1/p^n. I'm slightly reluctant to merge false `sorry`s on the basis that it's not clear whose job it will be to fix them.

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -0,0 +1,67 @@
...
+open Subgroup Set
+open scoped Pointwise
+
+namespace MeasureTheory
+variable {G α : Type*} [Group G] [MeasurableSpace G] [MeasurableMul G] [MeasurableSpace α]
+  {H K : Subgroup G}
+
+@[to_additive]
+instance : MeasurableMul H where
+  measurable_mul_const c := have := measurable_mul_const (c : G); sorry
```
kbuzzard, line 23: This one is just `by measurability`.

FLT/Mathlib/NumberTheory/Padics/MeasurableSpacePadics.lean
```diff
@@ -0,0 +1,52 @@
...
+noncomputable instance : MeasureSpace ℚ_[p] := ⟨addHaarMeasure Padic.unitBall_positiveCompact⟩
+
+instance : IsAddHaarMeasure (volume : Measure ℚ_[p]) := isAddHaarMeasure_addHaarMeasure _
+
+lemma Padic.volume_closedBall : volume {x : ℚ_[p] | ‖x‖ ≤ 1} = 1 := addHaarMeasure_self
+
+lemma PadicInt.volume_smul (x : nonZeroDivisors ℤ_[p]) :
+    volume (x • ⊤ : Set ℤ_[p]) = ‖x.1‖₊⁻¹ := sorry
+
+lemma Padic.volume_smul (x : ℚ_[p]ˣ) : volume (x • {y : ℚ_[p] | ‖y‖ ≤ 1}) = ‖(x:ℚ_[p])‖₊⁻¹ :=
```
kbuzzard, line 51:
```suggestion
lemma Padic.volume_smul (x : ℚ_[p]ˣ) : volume (x • {y : ℚ_[p] | ‖y‖ ≤ 1}) = ‖(x:ℚ_[p])‖₊ :=
```

FLT/Mathlib/NumberTheory/Padics/PadicIntegers.lean
```diff
@@ -0,0 +1,56 @@
...
+@[simp] lemma coe_inj {x y : ℤ_[p]} : (x : ℚ_[p]) = (y : ℚ_[p]) ↔ x = y := coe_injective.eq_iff
+
+noncomputable instance : Coe ℤ_[p]ˣ ℚ_[p]ˣ where coe := Units.map Coe.ringHom.toMonoidHom
+
+--TODO(javierlcontreras): Waiting for a new mathlib refactor. Z_p valuation should return nat, not int
+--lemma unit_ppown (a : ℤ_[p]) : ∃ u : ℤ_[p]ˣ, u * (p ^ a.valuation) = a := by
+--  sorry
+
+lemma index_ideal (a : ℤ_[p]) : (a • ⊤ : Ideal ℤ_[p]).toAddSubgroup.index = ‖a‖⁻¹ := by
+  sorry
```
kbuzzard, line 34: You sure you don't want to restrict to a != 0?

FLT/Mathlib/NumberTheory/Padics/PadicIntegers.lean
```diff
@@ -0,0 +1,56 @@
...
+
+lemma index_ideal_unit (u : ℤ_[p]ˣ) :
+    (u • ⊤ : Ideal ℤ_[p]).toAddSubgroup.index = ‖(u : ℤ_[p])‖⁻¹ := by
+  simp
+  sorry
```
kbuzzard, line 39: They're not just equal, they're both equal to 1. Are you sure you want this lemma? Wouldn't you be better off just proving they were both 1, if you definitely want this?

FLT/Mathlib/RingTheory/Norm/Defs.lean
```diff
@@ -0,0 +1,8 @@
+import Mathlib.RingTheory.Norm.Defs
+
+namespace LinearMap
+variable {M A B : Type*} [CommRing A] [CommRing B] [AddCommGroup M] [Algebra A B] [Module A M]
+  [Module B M] [IsScalarTower A B M] [Module.Finite A B] [Module.Finite B M]
```
kbuzzard, line 5:
```suggestion
  [Module B M] [IsScalarTower A B M] [Module.Finite A B] [Module.Finite B M]
  [Module.Free A B] [Module.Free B M]
```
otherwise it's likely to be false (`norm` is pathological without freeness and probably det is too).

FLT/Mathlib/LinearAlgebra/Determinant.lean
```diff
@@ -0,0 +1,8 @@
+import Mathlib.LinearAlgebra.Determinant
+
+namespace LinearMap
+variable {R : Type*} [CommRing R]
+
+variable (R) in
+@[simp] lemma det_mul (a : R) : (mul R R a).det = a := by
+  sorry
```
kbuzzard, line 8:
```suggestion
  classical
  rw [det_eq_det_toMatrix_of_finset (s := {1}) ⟨(Finsupp.LinearEquiv.finsuppUnique R R _).symm⟩]
  rw [Matrix.det_unique]
  change a * _ = a
  convert mul_one a
  simp
```

FLT/Mathlib/MeasureTheory/Group/Action.lean
YaelDillies: Oh really! `fun_prop` didn't work

FLT/Mathlib/RingTheory/Norm/Defs.lean
kbuzzard: I have added this file to `main` so I can make an issue of the `sorry`.

FLT/Mathlib/NumberTheory/Padics/MeasurableSpacePadics.lean
YaelDillies: Javier and I think the current statement is correct

YaelDillies: Nope, it's incorrect

FLT/Mathlib/NumberTheory/Padics/PadicIntegers.lean
YaelDillies: It's still true for `a = 0` thanks to junk values!

FLT/HaarMeasure/DistribHaarChar/Padic.lean
```diff
@@ -0,0 +1,106 @@
...
+
+This file computes `distribHaarChar` in the case of the actions of `ℤ_[p]ˣ` on `ℤ_[p]` and of
+`ℚ_[p]ˣ` on `ℚ_[p]`.
+
+This lets us know what `volume (x • s)` is in terms of `‖x‖` and `volume s`, when `x` is a
+p-adic/p-adic integer and `s` is a set of p-adics/p-adic integers.
+
+## Main declarations
+
+* `distribHaarChar_padic`: `distribHaarChar ℚ_[p]` is the usual p-adic norm on `ℚ_[p]`.
```
kbuzzard, line 21: It's the usual p-adic norm on Q_p^*. It's not defined at 0.

FLT/HaarMeasure/DistribHaarChar/Padic.lean
```diff
@@ -0,0 +1,106 @@
...
+* `distribHaarChar_padicInt`: `distribHaarChar ℤ_[p]` is the usual p-adic norm on `ℤ_[p]`.
```
kbuzzard, line 22: `distribHaarChar ℤ_[p]` is a character on Z_p^* and is the constant function 1. It's not even defined on all of Z_p. This docstring is quite misleading I think?

FLT/HaarMeasure/DistribHaarChar/Padic.lean
```diff
@@ -0,0 +1,106 @@
...
+
+@[simp] lemma PadicInt.volume_padicInt_smul (x : ℤ_[p]) (s : Set ℤ_[p]) :
+    volume (x • s) = ‖x‖₊ * volume s := by
+  simpa [-volume_padicInt_smul, ← image_coe_smul_set] using Padic.volume_padicInt_smul x ((↑) '' s)
+
+/-- The distributive Haar character of the action of `ℤ_[p]ˣ` on `ℤ_[p]` is the usual p-adic norm.
+
+This means that `volume (x • s) = ‖x‖ * volume s` for all `x : ℤ_[p]` and `s : Set ℤ_[p]`.
+See `PadicInt.volume_padicInt_smul` -/
+lemma distribHaarChar_padicInt (x : ℤ_[p]ˣ) : distribHaarChar ℤ_[p] x = ‖(x : ℤ_[p])‖₊ :=
```
kbuzzard, line 103: These are both equal to 1 so this lemma is very weird. Why not just prove it equals 1?

FLT/HaarMeasure/DistribHaarChar/RealComplex.lean
```diff
@@ -0,0 +1,77 @@
...
+    (isCompact_Icc.reProdIm isCompact_Icc).measure_ne_top ?_
+  -- The determinant of left multiplication by `z⁻¹` as a `ℝ`-linear map is `‖z‖₊ ^ (-2)`.
+  have key : ((LinearMap.mul ℂ ℂ z⁻¹).restrictScalars ℝ).det = ‖z.val‖₊ ^ (-2 : ℤ) := by
+    refine Complex.ofReal_injective ?_
+    rw [LinearMap.det_restrictScalars]
+    simp [Algebra.norm_complex_apply, normSq_eq_norm_sq, zpow_ofNat]
+  -- Massaging, we find the result.
+  have := addHaar_preimage_linearMap (E := ℂ) volume
+    (f := (LinearMap.mul ℂ ℂ z⁻¹).restrictScalars ℝ)
+  simp [key] at this
```
kbuzzard, line 65: Can you squeeze this?

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -0,0 +1,66 @@
...
+
+@[to_additive]
+instance (μ : Measure G) [μ.IsMulLeftInvariant] :
+    (μ.comap Subtype.val : Measure H).IsMulLeftInvariant where
+  map_mul_left_eq_self g := sorry -- https://github.com/ImperialCollegeLondon/FLT/issues/276
+
+@[to_additive]
+instance (μ : Measure G) [μ.IsMulRightInvariant] :
+    (μ.comap Subtype.val : Measure H).IsMulRightInvariant where
+  map_mul_right_eq_self g := sorry -- https://github.com/ImperialCollegeLondon/FLT/issues/276
```
kbuzzard, line 35: Issue #276 talks about a "measurable group". I don't know what that is, but I am surprised that inverse being measurable is not mentioned here. Is there a standard notion of measurable group?

FLT/Mathlib/NumberTheory/Padics/PadicIntegers.lean
```diff
@@ -0,0 +1,110 @@
...
+-- Yaël: Do we really want this as a coercion?
+noncomputable instance : Coe ℤ_[p]ˣ ℚ_[p]ˣ where coe := Units.map Coe.ringHom.toMonoidHom
+
+/-- `x • sℤ_[p]` has index `‖x‖⁻¹` in `sℤ_[p]`.
+
+Note that `sℤ_[p]` is the form `yℤ_[p]` for some `y : ℚ_[p]`, but this is syntactically less
+general. -/
+lemma smul_submoduleSpan_relindex_submoduleSpan (x : ℤ_[p]) (s : Set ℚ_[p]) :
+    (x • (Submodule.span ℤ_[p] s).toAddSubgroup).relindex (Submodule.span ℤ_[p] s).toAddSubgroup
+      = ‖x‖₊⁻¹ :=
```
kbuzzard, line 67: The only reference to `s` in this lemma is through  `(Submodule.span ℤ_[p] s)` so why not state it for `s : Submodule ℤ_[p] ℚ_[p]`? Then it will be applicable more widely. Same for next lemma.

kbuzzard (commented):
I think we're nearly there. Can you maybe not add any more stuff so we can get this over the line?

FLT/Mathlib/MeasureTheory/Group/Action.lean
YaelDillies: I didn't need `MeasurableInv`, but it can easily be stated

kbuzzard:
Thanks so much for your work on this!

Issue #255, opened 2024-12-01
An integer whose norm is < 1 must be 0


kbuzzard:
Exactly what it says on the tin. The `sorry` is in `NumberField/AdeleRing.lean`.

PR #260, opened 2024-12-01
chore: `|x| < 1` implies `x = 0` for `x : Int`

kim-em:
A bit ugly. There is a bug in `norm_cast` being exposed here that I couldn't minimize. We could also "clean up the mess" by making `omega` more robust against this failure.


kim-em:
Closes #255

kbuzzard:
Thanks!

PR #261, opened 2024-12-02
[BLUEPRINT]: Make the dependency graph colour scheme explicit

### LINKED ISSUE #69: understanding blueprint colourscheme
This code
```latex
\begin{lemma}
    \label{ZHat.e_not_in_Int}
    \lean{ZHat.e_not_in_Int}
    \uses{ZHat.e}
    \leanok
    The element $(e_N)_N$ of $\Zhat$ is not in $\Z$.
\end{lemma}
\begin{proof}\notready
    I don't have a proof of this but it doesn't look too bad; maybe I'm wrong
    and it's hard? We will not need it later on.
\end{proof}
```

produces (on FLT) an oval with a green border (fine) but a blue interior. I want that interior to be orange or red or something. 

In fact there seem to be two kinds of `\notready` proofs -- one is "I didn't write the LaTeX yet, feel free to write it for me" and the other is "this is a multi-year project, please see the discussion link". 
Maybe I want `\needsLaTeX` to be orange, and `\noplans` to be grey?

Can I also edit the legend to explain my personal colour system?

pitmonticone:
This PR makes the dependency graph’s color scheme explicit and easily editable.

Closes #69

### DIFF
--- a/blueprint/src/web.tex
+++ b/blueprint/src/web.tex
@@ -19,6 +19,15 @@
 \github{https://github.com/ImperialCollegeLondon/FLT/}
 \dochome{https://ImperialCollegeLondon.github.io/FLT/docs}
 
+% Configure the color scheme and legend description: \graphcolor{node_type}{color}{description}
+\graphcolor{stated}{green}{Green}              % Default values: stated, green, Green
+\graphcolor{can_state}{blue}{Blue}             % Default values: can_state, blue, Blue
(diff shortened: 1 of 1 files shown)


kbuzzard:
Thanks, and also thanks for your comments on the colour scheme. Right now "making the blueprint graph look nice" is low priority, I have realised that it's more worthwhile to focus on `sorry`s and use the blueprint to give more details for the harder ones. But it will no doubt become higher priority later on (in fact this will help, because I realised when I opened the issue that I didn't really understand things properly, and your comments on it mean that I now have a good reference).

PR #263, opened 2024-12-02
chore: rename `modularHaarChar` to `distribHaarChar`

YaelDillies:
This will avoid confusion with the modular character of a group, which is something else.

Also split off the definition of the `DomMulAct`-action on measures to a new file.


kbuzzard:
Thanks a lot!

PR #264, opened 2024-12-02
CI: cache more docs

YaelDillies:
The current list is very outdated.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -75,23 +75,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: |
-            .lake/build/doc/Aesop
-            .lake/build/doc/Batteries
-            .lake/build/doc/Cli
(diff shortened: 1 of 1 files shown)


kbuzzard:
@pitmonticone what do you think of this?

pitmonticone:
Hi @YaelDillies, thank you very much! You're right, the list is definitely outdated. 

What about saving to cache the whole `.lake/build/doc` without excluding any sub-contents? I'm doing in multiple projects (InfinityCosmos, Carleson, Equational) and it works perfectly fine without occupying too much memory.

YaelDillies:
Does it do the right thing with the project's own docs? If so, then yes that's definitely cleaner!

pitmonticone:
Unfortunately, I'm from mobile now so feel free to push it directly so I can review and merge it.

kbuzzard:
Thanks a lot!

PR #267, opened 2024-12-02
Add `ContinuousAlgEquiv`

### LINKED ISSUE #238: Continuous algebra equivalences
We don't have `ContinuousAlgEquiv` in mathlib as far as I can see: this is the structure whose terms are continuous R-algebra isomorphisms with continuous inverses. We do have `ContinuousLinearEquiv` though, so it's just a case of putting together this and `AlgEquiv`.

The reason we need it: we want to state that if L/K is a finite extension of number fields and v is a finite place of K, then the algebra isomorphism L tensor_K K_v = prod_{w|v} L_w is bicontinuous.

smmercuri:
Closes #238


kbuzzard:
Many thanks!

grunweg:
@smmercuri Would you like to PR the same file to mathlib? (You can ask me for review, if you like.)

PR #274, opened 2024-12-08
chore: fix sign in DistribHaarChar

kbuzzard:
#223 made me realise that the definition of  `DistribHaarChar` was off by a "multiplicative sign" i.e. an inverse (for example `lemma distribHaarChar_real (x : ℝˣ) : distribHaarChar ℝ x = ‖(x : ℝ)‖₊⁻¹ := ...` is proved in that PR, which was not what was supposed to happen) . The sign in the Lean currently doesn't agree with the sign in the docstring or the sign in the associated Zulip discussion. This PR changes the definition to make it correct.


YaelDillies:
Thanks! I will fix #223

PR #286, opened 2024-12-16
feat(RepresentationTheory): Adds Subrepresentations

javierlcontreras:
Co-authored-by: Yaël Dillies yael.dillies@gmail.com

### DIFF
--- a/FLT/Deformations/RepresentationTheory/Irreducible.lean
+++ b/FLT/Deformations/RepresentationTheory/Irreducible.lean
@@ -1,0 +1,30 @@
+import FLT.Deformations.RepresentationTheory.Subrepresentation
+import FLT.Mathlib.RepresentationTheory.Basic
+
+namespace Representation
+
+variable {G : Type*} [Group G]
+
+variable {k : Type*} [Field k]
+
+variable {W : Type*} [AddCommMonoid W] [Module k W]
+
+/-!
+  `IsIrreducible ρ` is the statement that a given Representation ρ is irreducible (also known as simple),
+  meaning that any subrepresentation must be either the full one (⊤) or zero (⊥)
+
+  This notion is only well behaved when the representation is over a field k. If it were defined over
+  a ring A with a nontrivial ideal J, the subrepresentation JW would often be a non trivial subrepresentation,
+  so ρ would rarely be irreducible.
+-/
+class IsIrreducible (ρ : Representation k G W) : Prop where
+  irreducible : IsSimpleOrder (Subrepresentation ρ)
+
+/-!
+  `IsAbsolutelyIrreducible ρ` states that a given Representation ρ over a field k
+  is absolutely irreducible, meaning that all the possible base change extensions are irreducible.
+-/
+class IsAbsolutelyIrreducible (ρ : Representation k G W) : Prop where
+  absolutelyIrreducible : ∀ k', ∀ _ : Field k', ∀ _ : Algebra k k', IsIrreducible (k' ⊗ᵣ' ρ)
+
+end Representation
--- a/FLT/Deformations/RepresentationTheory/Subrepresentation.lean
+++ b/FLT/Deformations/RepresentationTheory/Subrepresentation.lean
@@ -1,0 +1,74 @@
+import Mathlib.RepresentationTheory.Basic
+import FLT.Mathlib.LinearAlgebra.Span.Defs
+
+open Pointwise
+
+universe u
+
+variable {A : Type*} [CommRing A]
+
+variable {G : Type*} [Group G]
+
+variable {W : Type*} [AddCommMonoid W] [Module A W]
+
+variable {ρ : Representation A G W}
+
+variable (ρ) in
+structure Subrepresentation where
+  toSubmodule : Submodule A W
+  apply_mem_toSubmodule (g : G) ⦃v : W⦄ : v ∈ toSubmodule → ρ g v ∈ toSubmodule
+
+namespace Subrepresentation
+
+lemma toSubmodule_injective : Function.Injective (toSubmodule : Subrepresentation ρ → Submodule A W) := by
+  rintro ⟨_,_⟩
+  congr!
+
+instance : SetLike (Subrepresentation ρ) W where
+  coe ρ' := ρ'.toSubmodule
+  coe_injective' := SetLike.coe_injective.comp toSubmodule_injective
+
+def toRepresentation (ρ' : Subrepresentation ρ): Representation A G ρ'.toSubmodule where
+  toFun g := (ρ g).restrict (ρ'.apply_mem_toSubmodule g)
+  map_one' := by ext; simp
+  map_mul' x y := by ext; simp
+
+instance : Max (Subrepresentation ρ) where
+  max ρ₁ ρ₂ := .mk (ρ₁.toSubmodule ⊔ ρ₂.toSubmodule) <| by
+      simp only [Submodule.forall_mem_sup, map_add]
+      intro g x₁ hx₁ x₂ hx₂
+      exact Submodule.mem_sup.mpr
+        ⟨ρ g x₁, ρ₁.apply_mem_toSubmodule g hx₁, ρ g x₂, ρ₂.apply_mem_toSubmodule g hx₂, rfl⟩
+
+instance : Min (Subrepresentation ρ) where
+  min ρ₁ ρ₂ := .mk (ρ₁.toSubmodule ⊓ ρ₂.toSubmodule) <| by
+      simp only [Submodule.mem_inf, and_imp]
+      rintro g x hx₁ hx₂
+      exact ⟨ρ₁.apply_mem_toSubmodule g hx₁, ρ₂.apply_mem_toSubmodule g hx₂⟩
+
+
+@[simp, norm_cast]
+lemma coe_sup (ρ₁ ρ₂ : Subrepresentation ρ) : ↑(ρ₁ ⊔ ρ₂) = (ρ₁ : Set W) + (ρ₂ : Set W) :=
+  Submodule.coe_sup ρ₁.toSubmodule ρ₂.toSubmodule
+
+@[simp, norm_cast]
+lemma coe_inf (ρ₁ ρ₂ : Subrepresentation ρ) : ↑(ρ₁ ⊓ ρ₂) = (ρ₁ ∩ ρ₂ : Set W) := rfl
+
+@[simp]
+lemma toSubmodule_sup (ρ₁ ρ₂ : Subrepresentation ρ) :
+  (ρ₁ ⊔ ρ₂).toSubmodule = ρ₁.toSubmodule ⊔ ρ₂.toSubmodule := rfl
+
+@[simp]
+lemma toSubmodule_inf (ρ₁ ρ₂ : Subrepresentation ρ) :
+  (ρ₁ ⊓ ρ₂).toSubmodule = ρ₁.toSubmodule ⊓ ρ₂.toSubmodule := rfl
+
+instance : Lattice (Subrepresentation ρ) :=
+  toSubmodule_injective.lattice _ toSubmodule_sup toSubmodule_inf
+
+instance : BoundedOrder (Subrepresentation ρ) where
+  top := ⟨⊤, by simp⟩
+  le_top _ _ := by simp
+  bot := ⟨⊥, by simp⟩
+  bot_le _ _ := by simp +contextual
+
+end Subrepresentation
--- a/FLT/Mathlib/LinearAlgebra/Span/Defs.lean
+++ b/FLT/Mathlib/LinearAlgebra/Span/Defs.lean
@@ -1,0 +1,28 @@
+import Mathlib.LinearAlgebra.Span.Defs
+
+open Pointwise
+
+variable {R : Type*} [Semiring R]
+
+variable {M : Type*} [AddCommMonoid M] [Module R M]
+
+variable {p p' : Submodule R M}
+
+variable {P : M → Prop}
+
+namespace Submodule
+
+@[simp high]
+lemma forall_mem_sup : (∀ x ∈ p ⊔ p', P x) ↔ (∀ x₁ ∈ p, ∀ x₂ ∈ p', P (x₁ + x₂)) := by
+  simp [mem_sup]
+  aesop
+
+@[simp high]
+lemma exists_mem_sup : (∃ x ∈ p ⊔ p', P x) ↔ (∃ x₁ ∈ p, ∃ x₂ ∈ p', P (x₁ + x₂)) := by
+  simp [mem_sup]
+
+@[simp, norm_cast]
+lemma coe_sup' : ↑(p ⊔ p') = (p : Set M) + (p' : Set M) := by
+  simp [coe_sup]
+
+end Submodule
--- a/FLT/Mathlib/RepresentationTheory/Basic.lean
+++ b/FLT/Mathlib/RepresentationTheory/Basic.lean
@@ -1,0 +1,38 @@
+import Mathlib.LinearAlgebra.Matrix.GeneralLinearGroup.Defs
+import Mathlib.RepresentationTheory.Basic
+
+open LinearMap
+open scoped TensorProduct
+
+namespace Representation
+
+variable {R V G ι: Type*} [CommRing R] [AddCommMonoid V] [Module R V] [Module.Free R V]
+  [Module.Finite R V] [Group G] [DecidableEq ι] [Fintype ι]
+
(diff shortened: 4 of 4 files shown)


javierlcontreras:
@YaelDillies there is one remaining sorry at ```Subrepresentation.coe_sup``` that I could not solve, any tips?

YaelDillies:
The proof should be `toSubmodule_injective <| Submodule.coe_sup ..`

FLT/Deformations/RepresentationTheory/Subrepresentation.lean
```diff
@@ -0,0 +1,74 @@
...
+  coe_injective' := SetLike.coe_injective.comp toSubmodule_injective
+
+def toRepresentation (ρ' : Subrepresentation ρ): Representation A G ρ'.toSubmodule where
+  toFun g := (ρ g).restrict (ρ'.apply_mem_toSubmodule g)
+  map_one' := by ext; simp
+  map_mul' x y := by ext; simp
+
+instance : Max (Subrepresentation ρ) where
+  max ρ₁ ρ₂ := .mk (ρ₁.toSubmodule ⊔ ρ₂.toSubmodule) <| by
+      simp [Submodule.forall_mem_sup]
```
kbuzzard, line 38: Can you remove the nonterminal `simp`?

FLT/Deformations/RepresentationTheory/Subrepresentation.lean
```diff
@@ -0,0 +1,74 @@
...
+      intro g x₁ hx₁ x₂ hx₂
+      exact Submodule.mem_sup.mpr
+        ⟨ρ g x₁, ρ₁.apply_mem_toSubmodule g hx₁, ρ g x₂, ρ₂.apply_mem_toSubmodule g hx₂, rfl⟩
+
+instance : Min (Subrepresentation ρ) where
+  min ρ₁ ρ₂ := .mk (ρ₁.toSubmodule ⊓ ρ₂.toSubmodule) <| by
+      simp
```
kbuzzard, line 45: Ditto

FLT/Mathlib/RepresentationTheory/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+
+variable (ρ : Representation R G V) (𝓑 : Basis ι R V)
+
+omit [Module.Free R V] [Module.Finite R V] in
+@[simp]
+lemma comp_eq_mul (g h : G): ρ g ∘ₗ ρ h = ρ (g * h) := by
+  simp_all only [map_mul]
+  rfl
+
+noncomputable def GL_map_of_representation_of_basis
```
kbuzzard, line 20: Terms should start with a small letter.

FLT/Mathlib/RepresentationTheory/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+      val_inv := by rw [← toMatrix_comp, comp_eq_mul]; simp
+      inv_val := by rw [← toMatrix_comp, comp_eq_mul]; simp
+    }
+    map_one' := by aesop
+    map_mul' := by rintro x y; simp [LinearMap.toMatrix_mul]; norm_cast
+
+noncomputable def tprod' (R' : Type*) [CommRing R'] [Algebra R R'] (ρ : Representation R G V)
+    : Representation R G (R' ⊗[R] V) where
+  toFun g := TensorProduct.map .id (ρ g)
+  map_one' := by simp; rw [← LinearMap.one_eq_id, TensorProduct.map_one]
```
kbuzzard, line 34: Can you get rid of the non-terminal `simp`?

FLT/Deformations/RepresentationTheory/Subrepresentation.lean
```diff
@@ -0,0 +1,74 @@
...
+      intro g x hx₁ hx₂
+      exact ⟨ρ₁.apply_mem_toSubmodule g hx₁, ρ₂.apply_mem_toSubmodule g hx₂⟩
+
+
+@[simp, norm_cast]
+lemma coe_sup (ρ₁ ρ₂ : Subrepresentation ρ) : ↑(ρ₁ ⊔ ρ₂) = (ρ₁ : Set W) + (ρ₂ : Set W) := by
+  sorry
```
kbuzzard, line 52:
```suggestion
lemma coe_sup (ρ₁ ρ₂ : Subrepresentation ρ) : ↑(ρ₁ ⊔ ρ₂) = (ρ₁ : Set W) + (ρ₂ : Set W) :=
  Submodule.coe_sup ρ₁.toSubmodule ρ₂.toSubmodule
```
(found by noting that `rfl` didn't work and then trying `example (M N : Submodule A W) : M ⊔ N = (M : Set W) + N := by exact?`)

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,19 @@
...
+variable {A : Type*} [CommRing A]
+
+variable {W : Type*} [AddCommMonoid W] [Module A W]
+
+class Irreducible (ρ : Representation A G W) : Prop where
+  irreducible : IsSimpleOrder (Subrepresentation ρ)
+
+class AbsolutelyIrreducible (ρ : Representation A G W) : Prop where
+  absolutelyIrreducible : ∀ A', ∀ _ : CommRing A', ∀ _ : Algebra A A',
+    ∀ _ : Function.Injective (algebraMap A A'), Irreducible (A' ⊗ᵣ' ρ)
```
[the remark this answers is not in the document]
javierlcontreras: This makes sense, good catch.

FLT/Deformations/RepresentationTheory/Subrepresentation.lean
javierlcontreras: Are we happy with a non terminal "simp only"? I am guessing that the reason for asking for not having non terminal simps is to make proofs more robust to changes in the simp algorithm, but the "simp only"  case is some middle point which might be fine, depending on the community's preference

kbuzzard: Yes definitely! The rules are that after `simp` there are very few things which are allowed to happen (e.g. `ring`) but most of the time `simp` is supposed to be a closing tactic. Just replace `simp` with `simp?` and then the corresponding `simp only` (or see if you can remove the `simp` completely if you like).

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,19 @@
...
+variable {k : Type*} [Field k]
+
+variable {W : Type*} [AddCommMonoid W] [Module k W]
+
+class Irreducible (ρ : Representation k G W) : Prop where
+  irreducible : IsSimpleOrder (Subrepresentation ρ)
+
+class AbsolutelyIrreducible (ρ : Representation k G W) : Prop where
+  absolutelyIrreducible : ∀ k', ∀ _ : Field k', ∀ _ : Algebra k k',
+    ∀ _ : Function.Injective (algebraMap k k'), Irreducible (k' ⊗ᵣ' ρ)
```
kbuzzard, line 17: This definition is now weird, because if k has zero divisors then there will be no fields at all which admit an injective map from k, and thus every k-representation will be vacuously absolutely irreducible. Why not drop injectivity? It will be automatic in all the cases in which one wants to apply it, namely when k is a field, and it simplifies the definition to drop it. The class should also be called `IsAbsolutelyIrreducible` because it's a Prop, it should have a docstring, and the docstring should explain that it gives the right answer when `k` is a field and otherwise should be used at your own risk.

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Representation
+
+variable {G : Type*} [Group G]
+
+variable {k : Type*} [Field k]
+
+variable {W : Type*} [AddCommMonoid W] [Module k W]
+
+class Irreducible (ρ : Representation k G W) : Prop where
```
kbuzzard, line 12: This is a Prop so should be `IsIrreducible`.

FLT/Mathlib/LinearAlgebra/Span/Defs.lean
```diff
@@ -0,0 +1,28 @@
...
+variable {p p' : Submodule R M}
+
+variable {P : M → Prop}
+
+namespace Submodule
+
+@[simp high]
+lemma forall_mem_sup : (∀ x ∈ p ⊔ p', P x) ↔ (∀ x₁ ∈ p, ∀ x₂ ∈ p', P (x₁ + x₂)) := by
+  simp [mem_sup]
+  aesop
```
kbuzzard, line 18: Here is a (rare) situation where a non-terminal simp is OK. Because aesop uses simp (it uses the kitchen sink) so if a change in simp breaks this proof then it was probably going to break anyway. After a non-terminal simp you can only have very strong tactics, basically.

FLT/Mathlib/RepresentationTheory/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+variable {R V G ι: Type*} [CommRing R] [AddCommMonoid V] [Module R V] [Module.Free R V]
+  [Module.Finite R V] [Group G] [DecidableEq ι] [Fintype ι]
+
+variable (ρ : Representation R G V) (𝓑 : Basis ι R V)
+
+omit [Module.Free R V] [Module.Finite R V] in
+@[simp]
+lemma comp_eq_mul (g h : G): ρ g ∘ₗ ρ h = ρ (g * h) := by
+  simp_all only [map_mul]
+  rfl
```
kbuzzard, line 18: The fact that you needed `rfl` after `simp` indicates that there's something wrong here. The goal after `simp_all only [map_mul]` is
```
⊢ ρ g ∘ₗ ρ h = ρ g * ρ h
```
and indeed this is the goal after `simp` (you don't need `simp_all` here, there's nothing in the local context which is helping). The fact that `simp` turns the RHS into `ρ g * ρ h` means that the community has decided that `ρ (g * h)` is not in simp normal form, and that the simp normal form for this expression is `ρ g * ρ h`. So making `comp_eq_mul` into a `simp` lemma is a dubious choice because you are simplifying something into a form which is not the simplest form. Why not
```
@[simp]
lemma comp_def (g h : G): ρ g ∘ₗ ρ h = ρ g * ρ h := rfl
```
? That way you're putting the equation into the form which the community (or at least whoever wrote `map_mul` and tagged it `@[simp]` ) has decided is the simp normal form.

FLT/Mathlib/RepresentationTheory/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+
+noncomputable def gl_map_of_representation_of_basis
```
kbuzzard, line 20: I can't believe that this is what the naming convention says this map is called (but I'm not very good at the naming convention). At the very least you can remove `of_representation` because you're in the Representation namespace right now.

FLT/Mathlib/RepresentationTheory/Basic.lean
```diff
@@ -0,0 +1,40 @@
...
+      val := LinearMap.toMatrix 𝓑 𝓑 (ρ g)
+      inv := LinearMap.toMatrix 𝓑 𝓑 (ρ g⁻¹)
+      val_inv := by rw [← toMatrix_comp, comp_eq_mul]; simp
+      inv_val := by rw [← toMatrix_comp, comp_eq_mul]; simp
+    }
+    map_one' := by aesop
+    map_mul' := by rintro x y; simp [LinearMap.toMatrix_mul]; norm_cast
+
+noncomputable def tprod' (R' : Type*) [CommRing R'] [Algebra R R'] (ρ : Representation R G V)
+    : Representation R G (R' ⊗[R] V) where
```
kbuzzard, line 32: Did you really mean to make R' (x)_R V an R-representation rather than R'-representation here? If you meant it to be an R'-representation then it could be called baseChange or something like that. If you did want it to be an R-representation then my instinct is to make baseChange anyway, and also restrictScalars from R'-reps to R-reps if R' is an R-algebra, and get tprod' by composing.

kbuzzard (commented):
Here are some more comments. Can you click on "resolve conversation" for comments which you've dealt with (e.g. if you've applied suggestions which I've made). That way it's much easier to see what has been dealt with and what remains to be done.

FLT/Deformations/RepresentationTheory/Irreducible.lean
javierlcontreras: As of right now, the definition needs k to be a field right? So I don't understand the "and otherwise should be used at your own risk." part

javierlcontreras: I do see how injectivity is now free, happy to drop that

javierlcontreras: Extra question: How could k have zero divisors if k is a field?

kbuzzard: Oh apologies! I had missed that `k` was already a field, so some of what I said above is nonsense (I'd just read through the subrepresentation file, where k was a commring). Sorry for creating confusion. My comment should have just said "drop injectivity, this is automatic for fields". The definition is fine.

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,34 @@
...
+namespace Representation
+
+variable {G : Type*} [Group G]
+
+variable {k : Type*} [Field k]
+
+variable {W : Type*} [AddCommMonoid W] [Module k W]
+
+/-!
+  IsIrreducible predicates that a given Representation ρ is irreducible (also known as simple),
```
kbuzzard, line 13:
```suggestion
  `IsIrreducible ρ` is the statement that a given representation ρ is irreducible (also known as simple),
```
My instinct is to make things easier for mathematicians to read, and I don't know if they know what "predicates" means.

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,34 @@
...
+  meaning that any subrepresentation must be either the full one (⊤) or zero (⊥)
+
+  This notion is only well behaved when the representation is over a field k. If it were defined over
+  a ring A with a nontrivial ideal J, the subrepresentation JW would be a non trivial subrepresentation,
```
kbuzzard, line 17:
```suggestion
  a ring A with a nontrivial ideal J, the subrepresentation JW would often be a non trivial subrepresentation,
```
If W was A/J then this trick doesn't work.

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,34 @@
...
+  so ρ would never be irreducible.
```
kbuzzard, line 18:
```suggestion
  so ρ would rarely be irreducible.
```

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,34 @@
...
+class IsIrreducible (ρ : Representation k G W) : Prop where
+  irreducible : IsSimpleOrder (Subrepresentation ρ)
+
+/-!
+  IsAbsolutelyIrreducible predicates that a given Representation ρ over a field k
+  is absolutely irreducible, meaning that all the possible base change extensions are irreducible.
+
+  This notion is only well behaved when the representation is over a field k. If it were defined over
+  a ring A with a nontrivial ideal J, the subrepresentation JW would be a non trivial subrepresentation,
+  so ρ would never be absolutely irreducible.
```
kbuzzard, line 29:
```suggestion
```
Sorry, this was my misunderstanding.

FLT/Deformations/RepresentationTheory/Irreducible.lean
```diff
@@ -0,0 +1,34 @@
...
+
+  This notion is only well behaved when the representation is over a field k. If it were defined over
+  a ring A with a nontrivial ideal J, the subrepresentation JW would be a non trivial subrepresentation,
+  so ρ would never be irreducible.
+-/
+class IsIrreducible (ρ : Representation k G W) : Prop where
+  irreducible : IsSimpleOrder (Subrepresentation ρ)
+
+/-!
+  IsAbsolutelyIrreducible predicates that a given Representation ρ over a field k
```
kbuzzard, line 24:
```suggestion
  `IsAbsolutelyIrreducible ρ` states that a given representation ρ over a field k
```

kbuzzard (approved):
Modulo docstrings this is good to go. Thanks a lot!

kbuzzard:
Oh sorry, you dealt with all my comments but didn't resolve the conversations so I only just noticed!

PR #304, opened 2025-01-05
Revert "Updates available and ready to merge"

kbuzzard:
Reverts ImperialCollegeLondon/FLT#303

Unfortunately I'm going to have to revert the recent bump because `lake exe cache get` no longer works, meaning that I can't get cache so can't in practice develop the project.


kbuzzard:
cc @pitmonticone

Issue #307, opened 2025-01-06
(Integers of Q) ~-> Z -> Q triangle commutes


kbuzzard:
This is the sorry in `FLT.Mathlib.NumberTheory.NumberField.Basic`. It says that Mathlib's isomorphism `𝓞 ℚ ≃+* ℤ` commutes with the obvious maps from both sides into `ℚ`. 

```
import Mathlib.NumberTheory.NumberField.Basic

open scoped NumberField

theorem Rat.ringOfIntegersEquiv_eq_algebraMap (z : 𝓞 ℚ) :
    (Rat.ringOfIntegersEquiv z : ℚ) = algebraMap (𝓞 ℚ) ℚ z := by
  sorry
```

kbuzzard:
claim

kbuzzard:
propose #308

PR #308, opened 2025-01-06
feat: progress towards Rat.AdeleRing.zero_discrete

kbuzzard:
We fill in a sorry in the proof that there's an open subset of the adeles of Q whose intersection with Q is {0}.


kbuzzard:
This closes #307 and #254

PR #310, opened 2025-01-11
Issue #259: A_K/K is compact if A_Q/Q is compact

### LINKED ISSUE #259: A_K / K is compact for K a number field
Assuming that A_Q / Q is compact and K tensor A_Q is A_K, we can deduce that A_K / K is compact. The details are [in the blueprint](https://imperialcollegelondon.github.io/FLT/blueprint/Adele_miniproject.html#NumberField.AdeleRing.cocompact). 

The sorry is in `NumberField/AdeleRing.lean`.

smmercuri:
Closes #259


kbuzzard:
I've bumped main in the FLT project because we had been stuck on 4.15 for technical reasons; this created conflicts in this PR but I've fixed them all in [this commit](https://github.com/ImperialCollegeLondon/FLT/commit/55ccbfd6df0c9ec8777c3543a06898be71ecaafc).

FLT/Mathlib/RingTheory/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,12 @@
...
+
+theorem Algebra.TensorProduct.piScalarRight_symm_apply_of_algebraMap (R S N ι : Type*)
+    [CommSemiring R] [CommSemiring S] [Algebra R S] [Semiring N] [Algebra R N] [Algebra S N]
+    [IsScalarTower R S N] [Fintype ι] [DecidableEq ι] (x : ι → R) :
+    (TensorProduct.piScalarRight R S N ι).symm (fun i => algebraMap _ _ (x i)) =
+      1 ⊗ₜ[R] (∑ i, Pi.single i (x i)) := by
+  simp only [_root_.TensorProduct.piScalarRight_symm_apply, LinearMap.lsum_apply,
+    LinearMap.coeFn_sum, LinearMap.coe_comp, LinearMap.coe_mk, AddHom.coe_mk, LinearMap.coe_proj,
+    Finset.sum_apply, Function.comp_apply, Function.eval, Algebra.algebraMap_eq_smul_one,
+    TensorProduct.smul_tmul, ← TensorProduct.tmul_sum, ← Pi.single_smul, smul_eq_mul, mul_one]
```
kbuzzard, line 12:
```suggestion
  simp [LinearEquiv.symm_apply_eq, algebraMap_eq_smul_one]
```

FLT/Mathlib/LinearAlgebra/TensorProduct/Pi.lean
```diff
@@ -0,0 +1,10 @@
+import Mathlib.LinearAlgebra.TensorProduct.Pi
+
+theorem TensorProduct.piScalarRight_symm_apply (R S N ι : Type*) [CommSemiring R]
```
kbuzzard, line 3: I don't think we need this theorem. See my suggestion below for avoiding it the only time you use it.

FLT/NumberField/AdeleRing.lean
```diff
@@ -18,6 +23,141 @@ end LocallyCompact
 
 section BaseChange
 
+namespace NumberField.AdeleRing
+
+variable (K L : Type*) [Field K] [Field L] [NumberField K] [NumberField L] [Algebra K L]
+
+noncomputable instance : Algebra K (NumberField.AdeleRing L) :=
+  RingHom.toAlgebra <| (algebraMap L (NumberField.AdeleRing L)).comp <| algebraMap K L
```
kbuzzard, line 31:
```suggestion
noncomputable instance : Algebra K (NumberField.AdeleRing (𝓞 L) L) :=
  Algebra.compHom _ (algebraMap K L)
```
I think this is better because if K=L then you might get a diamond with your approach. Indeed this suggestion breaks a proof in a good way later on (it finishes early)

FLT/NumberField/AdeleRing.lean
```diff
@@ -18,6 +23,141 @@ end LocallyCompact
...
+    TopologicalSpace (NumberField.AdeleRing K ⊗[K] L) :=
+  letI := TopologicalSpace.induced (algebraMap K (AdeleRing K)) inferInstance
+  moduleTopology K _
+```
+However it is not clear to me how the inverse function is of `tensorProductLinearEquivPi`
+is continuous in that case. Additionally,
+https://math.mit.edu/classes/18.785/2017fa/LectureNotes25.pdf (just above Prop 25.10)
+for an informal source where the tensor product is given the product topology. Maybe they
+coincide!
+-/
```
kbuzzard, line 71: Yes, the module topology on a finite free R-module coincides with the product topology for any choice of basis (and in particular the topology on a finite free thing is independent of choice of basis). This should now all be in mathlib.

kbuzzard: Note to self; I haven't read the rest of this file.

FLT/NumberField/AdeleRing.lean
```diff
@@ -18,6 +23,141 @@ end LocallyCompact
...
+variable (K L)
+
+/-
+smmercuri : The tensor product here is in a different order to the one
+appearing in the finite adele ring base change, where `L ⊗[K] FiniteAdeleRing A K`
+is used. Is there a preference between these orderings? One benefit
+to using `AdeleRing K ⊗[K] L` is that it automatically gets
+a `K` algebra instance via the instance `Algebra.TensorProduct.leftAlgebra`
+(while `rightAlgebra` is a def), but maybe there are other reasons to
+prefer the `rightAlgebra` set up.
```
kbuzzard, line 101: Yeah :-/ Of course the benefit of using the other order is that you automatically get an `L`-algebra instance. I am still really unclear about the best way to proceed here. When doing automorphic forms it's really normal to have G(Q) \ G(A) and so I've been putting number fields on the left in general but that was my only logic.

FLT/NumberField/AdeleRing.lean
```diff
@@ -18,6 +23,141 @@ end LocallyCompact
...
+-/
+def baseChange : (AdeleRing K ⊗[K] L) ≃A[K] AdeleRing L := sorry
```
kbuzzard, line 103: Is there a way we can write down the data and only sorry the proofs here?

FLT/NumberField/AdeleRing.lean
```diff
@@ -18,6 +23,141 @@ end LocallyCompact
...
+
+variable {L}
+
+theorem baseChange_tsum_apply_right (l : L) :
+  baseChange K L (1 ⊗ₜ[K] l) = algebraMap L (AdeleRing L) l := sorry
```
kbuzzard, line 108: because then this theorem would be provable.

FLT/NumberField/AdeleRing.lean
erdOne: I think you should write `L ⊗ AdeleRing K` and make it an `L`-algebra homomorphism.

smmercuri:
Sorry, I somehow managed to close this PR without meaning to! The new one is open here #315 

It includes some fixes to comments here, but some still remain so it's WIP atm.

FLT/NumberField/AdeleRing.lean
smmercuri: I imagine this result will follow from some results that combine the two base changes for the infinite and finite adele ring. I can work on that in this PR as well.

FLT/Mathlib/LinearAlgebra/TensorProduct/Pi.lean
smmercuri: Nice, this has been removed in #315

PR #312, opened 2025-01-16
The continuous algebra homomorphism from `v.Completion` to `w.Completion` whenever the infinite place `w` extends `v`

smmercuri:


FLT/NumberField/Completion.lean
```diff
@@ -0,0 +1,83 @@
...
+  simp only [UniformSpace.Completion.algebraMap_def, algebraMap_coe _]
+  rfl
+
+instance {wv : v.ExtensionPlace L} : IsScalarTower K v.Completion wv.1.Completion :=
+  IsScalarTower.of_algebraMap_eq (algebraMap_comp wv)
+
+open UniformSpace.Completion in
+def extensionPlaceContinuousAlgHom (wv : v.ExtensionPlace L) :
+    v.Completion →A[v.Completion] wv.1.Completion where
+  __ := mapOfComp wv.abs_comp
```
kbuzzard, line 50: Why not just
```suggestion
  __ := algebraMap v.Completion wv.1.Completion
```
?

More generally I feel like this should be somehow coming from from a general function which constructs this from instances `Algebra v.Completion wv.1.Completion` and `ContinuousSMul v.Completion wv.1.Completion`.

FLT/NumberField/InfiniteAdeleRing.lean
```diff
@@ -1,18 +1,27 @@
...
 
-variable [Algebra K (InfiniteAdeleRing L)] [IsScalarTower K L (InfiniteAdeleRing L)]
+open scoped Classical TensorProduct
 
--- TODO should be →A[K]
-/-- The canonical map from the infinite adeles of K to the infinite adeles of L -/
-def NumberField.InfiniteAdeleRing.baseChange :
-    InfiniteAdeleRing K →ₐ[K] InfiniteAdeleRing L :=
-  sorry
+noncomputable local instance : Algebra K (InfiniteAdeleRing L) := Pi.algebra _ _
```
kbuzzard, line 11: I'm beginning to wonder if this is a dubious design decision. In the finite adele case I'm moving away from these instances and instead defining maps `FiniteAdeleRing A K -> FiniteAdeleRing B L` which "live above" `K -> L` via "semialgebra maps". What do you think?

kbuzzard:
I've fixed the conflicts in branch `InfinitePlaceAlgHom` in `upstream` so if you want to merge that then all that's left is the trivial comments above, which I could deal with but I thought I'd let you look at them at least rather than just edit your work. Note that ContinuousAlgEquiv was merged into mathlib a while ago.

FLT/Mathlib/Topology/Algebra/ContinuousAlgEquiv.lean
```diff
@@ -267,4 +268,67 @@ theorem _root_.AlgEquiv.isUniformEmbedding {E₁ E₂ : Type*} [UniformSpace E
...
 
+variable {ι α β : Type*}
+
+@[simps!]
+def piCurry (S : Type*) [CommSemiring S] {Y : ι → Type*}
+    (α : (i : ι) → Y i → Type*) [(i : ι) → (y : Y i) → Semiring (α i y)]
+    [(i : ι) → (y : Y i) → Algebra S (α i y)] [(i : ι) → (y : Y i) → TopologicalSpace (α i y)] :
+    ((i : Sigma Y) → α i.1 i.2) ≃A[S] ((i : ι) → (y : Y i) → α i y) where
+  toAlgEquiv := AlgEquiv.piCurry S α
+  continuous_toFun := continuous_pi (fun _ => continuous_pi <| fun _ => continuous_apply _)
```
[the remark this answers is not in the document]
smmercuri: Yeah somehow `fun_prop` is not able to unpack these. I would have expected it to work

FLT/NumberField/InfiniteAdeleRing.lean
smmercuri: Yes I think semialgebra maps are better. I've updated to use those throughout in this PR and it's much tidier. This is now analogous to `ProdAdicCompletions.baseChange` so I've abstracted that to `Pi.semalgHomPi`

FLT/NumberField/Completion.lean
smmercuri: Agreed, although this has now been replaced by semialg homs

smmercuri:
Thanks! After updating to use semialgebra maps, there are some unused results/definitions in this PR (basically all the `(Continuous)AlgEquiv.piCongrLeft`s and `piCurry`'s), wondering whether to remove or keep them?

kbuzzard:
I'm happy to get rid of them (or just PR them to mathlib if you think they're worth keeping)

smmercuri:
Removed them! This is now ready for review again (I don't think I can edit labels to remove awaiting-author)

PR #315, opened 2025-01-20
Issue #259: A_K/K is compact if A_Q/Q is compact

### LINKED ISSUE #259: A_K / K is compact for K a number field

smmercuri:
Closes #259


kbuzzard:
Is this still WIP?

smmercuri:
This one is still WIP although nearly there, I'll let you know when it's ready for review again

smmercuri:
This is now ready to review again

FLT/Mathlib/Algebra/Algebra/Tower.lean
```diff
@@ -0,0 +1,11 @@
+import Mathlib.Algebra.Algebra.Tower
+import Mathlib.RingTheory.AlgebraTower
+
+def AlgEquiv.extendScalars {A C D : Type*} (B : Type*) [CommSemiring A] [CommSemiring C]
+    [CommSemiring D] [Algebra A C] [Algebra A D] [CommSemiring B] [Algebra A B] [Algebra B C]
+    [IsScalarTower A B C] (f : C ≃ₐ[A] D) :
+    letI := (f.toAlgHom.restrictDomain B).toRingHom.toAlgebra
+    C ≃ₐ[B] D where
```
kbuzzard, line 8: Is this def actually usable? If `D` is already a `B`-algebra then this will cause a diamond perhaps. But maybe it's fine?

FLT/Mathlib/Topology/Algebra/Group/Quotient.lean
```diff
@@ -0,0 +1,13 @@
...
+import FLT.Mathlib.Topology.Algebra.Module.Quotient
+import FLT.Mathlib.Topology.Algebra.Module.Equiv
+
+def QuotientAddGroup.continuousAddEquiv (G H : Type*) [AddCommGroup G] [AddCommGroup H] [TopologicalSpace G]
+    [TopologicalSpace H] (G' : AddSubgroup G) (H' : AddSubgroup H) [G'.Normal] [H'.Normal]
+    (e : G ≃ₜ+ H) (h : AddSubgroup.map e G' = H') :
+    G ⧸ G' ≃ₜ+ H ⧸ H' :=
+  (Submodule.Quotient.continuousLinearEquiv _ _ (AddSubgroup.toIntSubmodule G')
+    (AddSubgroup.toIntSubmodule H') e.toIntContinuousLinearEquiv
+      (congrArg AddSubgroup.toIntSubmodule h)).toContinuousAddEquiv
```
kbuzzard, line 13: Probably all his should be done for `CommGroup` and then `@[to_additive]d`.

FLT/NumberField/AdeleRing.lean
```diff
@@ -98,54 +108,142 @@ theorem Rat.AdeleRing.discrete : ∀ q : ℚ, ∃ U : Set (AdeleRing (𝓞 ℚ)
...
+
+variable {K L}
+
+open TensorProduct.AlgebraTensorModule in
+theorem piEquiv_apply_of_algebraMap
+    {x : Fin (Module.finrank K L) → 𝔸 K}
+    {y : Fin (Module.finrank K L) → K}
+    (h : ∀ i, algebraMap K (𝔸 K) (y i) = x i) :
+    piEquiv K L x = algebraMap L _ (Module.Finite.equivPi _ _ |>.symm y) := by
+  simp [← funext h]
```
kbuzzard, line 194: Can you squeeze the simp?

kbuzzard (approved):
Thanks a lot! My comments are only trivial so I'll just merge and fix them myself. You should also update the LaTeX.

FLT/Mathlib/Topology/Algebra/Group/Quotient.lean
kbuzzard: Also, is there a reason why you went for `map e G' = H`? I would have thought that `comap e H' = G'` was easier to work with, because it has better definitional properties (the definition of `map` uses `image` so has an `exists` in, but the definition of `comap` uses `preimage` so doesn't).

FLT/Mathlib/Algebra/Algebra/Tower.lean
smmercuri: Yeah you have to be careful with this one. In this PR it's used in
```
-- `𝔸 K ⊗[K] L ≃ₗ[𝔸 K] L ⊗[K] 𝔸 K`
let comm := (Algebra.TensorProduct.comm K (𝔸 K) L).extendScalars (𝔸 K) |>.toLinearEquiv
```
Which works because `Algebra.TensorProduct.rightAlgebra` on the RHS defeqs to `Algebra.TensorProduct.includeRight.toRingHom.toAlgebra` and `includeRight` is `TensorProduct.comm.restrictDomain 𝔸 K`. But the other way around
```
let comm := (Algebra.TensorProduct.comm K L (𝔸 K)).extendScalars (𝔸 K) |>.toLinearEquiv
```
causes issues. Note that `AlgHom.extendScalars` in mathlib is the same in this respect. Maybe the right way to state this is to add `[Algebra B D]` and some kind of `CompatibleSMul` instance?

FLT/Mathlib/Topology/Algebra/Group/Quotient.lean
smmercuri: No particular reason other than using [`Submodule.Quotient.equiv`](https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/Quotient/Basic.html#Submodule.Quotient.equiv) as a model. It sounds like `comap` would be better!

PR #319, opened 2025-01-24
Adapt to the new doc-gen setup

YaelDillies:
Instead of making `FLT` depend on `doc-gen4` conditionally, have a new private Lean project `docbuild` that depends on `FLT` and `doc-gen4`.

The only change on the user side is that one needs to run `scripts/bump.sh` instead of `scripts/update_mathlib.sh`. I have renamed the script because it does not just update mathlib, and the new name is shorter.


pitmonticone:
Hi @YaelDillies, thanks a lot for your PR! 

The following config files are unnecessary:
- `docbuild/lean-toolchain`
- `docbuild/lake-manifest.json`

Everything should be configured and hosted by the workflow runner.

Could you please remove those files and change the "Build project API documentation" job step as follows?

```
      - name: Build project API documentation
        working-directory: docbuild
        run: |
          MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update FLT || true
          ~/.elan/bin/lake build FLT:docs
```

YaelDillies:
I'm not sure I understand your use of `lake update FLT`. `docbuild` should follow `doc-gen4`, not `FLT`. If ever there are conflicting toolchains between  `doc-gen4` and `FLT`, the correct one to follow is `doc-gen4`.

pitmonticone:
I followed the doc-gen4 [README](https://github.com/leanprover/doc-gen4): 

> If your parent project has dependencies you want to run lake update YourLibraryName within docbuild whenever you update the dependencies of your parent project.

YaelDillies:
Hmm, I find that confusing, but hopefully the README is correct.

Other question: If you do not want to make `docbuild` a Lean project that's functional locally, then why even have it at all? Why can't `docbuild/lakefile.toml` be created on the fly by CI?

pitmonticone:
> If you do not want to make docbuild a Lean project that's functional locally, then why even have it at all? Why can't docbuild/lakefile.toml be created on the fly by CI?

Well, you can make it functional right away with a quick `MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update FLT`. However, I might be biased because I have never found much use for local documentation.

pitmonticone (approved):
LGTM

kim-em:
Yes, thank you!

kbuzzard:
OK let's give it a go!

PR #321, opened 2025-01-26
`AbsoluteValue` API needed to prove weak approximation

smmercuri:


kbuzzard:
I wonder if there is some stuff from the recent proof of Ostrowski's theorem which you could have used here? Still, thanks a lot!

PR #331, opened 2025-02-02
Connect the Fujisaki lemmas in the blueprint

Ruben-VandeVelde:


kbuzzard:
Thanks! Right now my main job is to go through these miniprojects (adeles, haar characters, Fukisaki, quatertion algebras) and really turn them into something which others can work on. I've spent quite some time recently getting all the LaTeX down but haven't started thinking about gluing everything together blueprint style.

PR #334, opened 2025-02-05
CI: generate doc-building repository in CI

YaelDillies:
* Don't have `docbuild` lying around always. Currently, the toolchain it gets updated to is independent of the toolchain FLT is on, causing issue every first few days of the month. Now, the toolchain with which the docs are built will always be the one from FLT. No one can accidentally introduce a toolchain mismatch error in the repo by modifying/forgetting to modify the pinned toolchain in `docbuild/lakefile.toml` anymore.
* Add `scripts/build_docs.sh` to automate building the docs. This script creates a temporary `docbuild` repo, generates the docs, then deletes the temporary repo again.
* Remove `scripts/bump.sh`. Now that `docbuild` doesn't exist outside of CI/building the docs locally with `build_docs.sh`, instructions to update the repo are simply "Run `lake update`".


scripts/build_docs.sh
```diff
@@ -0,0 +1,45 @@
+# Buid HTML documentation for FLT
```
pitmonticone, line 1: Small typo: 
```suggestion
# Build HTML documentation for FLT
```

pitmonticone (approved):

scripts/build_docs.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Equip docbuild with the template lakefile.toml
+template > docbuild/lakefile.toml
+
+# Substitute the toolchain from lean-toolchain into docbuild/lakefile.toml
+sed -i s/TOOLCHAIN/`grep -oP 'v4\..*' lean-toolchain`/ docbuild/lakefile.toml
+
+# Initialise docbuild as a Lean project
+cd docbuild
+~/.elan/bin/lake update FLT
```
YaelDillies, line 34: Pietro, if you want to keep `MATHLIB_NO_CACHE_ON_UPDATE`, it's here:
```suggestion
MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update FLT
```
But do you want it? What does it even do?

pitmonticone: Unfortunately it's necessary to avoid `lake exe cache get` to run after the update and make the docbuild setup work.

Se [here](https://leanprover.zulipchat.com/#narrow/stream/270676-lean4/topic/New.20recommended.20default.20setup.20for.20doc-gen4) for further details.

pitmonticone: But wait. Did the workflow run without any issues?

YaelDillies: Ah right, it just gets rid of a non-blocking error message. I guess we can keep it

pitmonticone: Yes, agreed.

pitmonticone (approved):

kbuzzard:
Thanks a lot!

PR #337, opened 2025-02-06
chore: bump to stable `v4.16.0`

pitmonticone:
- [x] Upgrade deprecated lemmas and notation
- [x] Fix errors


scripts/bump.sh
YaelDillies, line 1: This change would be better as a standalone PR. Alternatively, drop it, since I am removing the script altogether in #334

pitmonticone: Yes, it makes sense.

FLT/HaarMeasure/DistribHaarChar/RealComplex.lean
```diff
@@ -31,7 +31,8 @@ open Real Complex MeasureTheory Measure Set
 open scoped Pointwise
 
 lemma Real.volume_real_smul (x : ℝ) (s : Set ℝ) : volume (x • s) = ‖x‖₊ * volume s := by
-  simp [← ennnorm_eq_ofReal_abs]
+  simp [← enorm_eq_ofReal_abs]
+  rfl
```
Ruben-VandeVelde, line 35:
```suggestion
  simp [← enorm_eq_ofReal_abs, enorm_eq_nnnorm]
```

Ruben-VandeVelde (approved):

PR #338, opened 2025-02-07
Adds inverse limits of Modules, Rings, Groups and Representations

javierlcontreras:
Needed for the formalization of existence of the Universal Deformation Ring (following Smit&Lenstra). 

This is not Mathlib-grade code, I am guessing (and @YaelDillies agreed) that both Representation and InverseLimit should be done categorically, but at the moment the CategoryTheory API is hard (to me at least). I implemented this to unblock the R project.


FLT/Deformations/Algebra/InverseLimit.lean
```diff
@@ -0,0 +1,218 @@
...
+
+open Submodule
+
+variable {ι : Type*} [Preorder ι] [DecidableEq ι]
+variable {obj : ι → Type*}
+
+namespace Module
+
+variable {R : Type*} [Ring R]
+variable [addCommGroupObj : ∀ {i : ι}, AddCommGroup (obj i)] [∀ i, Module R (obj i)]
```
YaelDillies, line 32:
```suggestion
variable [addCommGroupObj : ∀ i, AddCommGroup (obj i)] [∀ i, Module R (obj i)]
```

FLT/Deformations/Algebra/InverseLimit.lean
```diff
@@ -0,0 +1,218 @@
...
+  carrier := { a : Π i : ι, obj i |
+    ∀ (i j : _) (h : i ≤ j), func h (a j) = a i }
+  add_mem' := by aesop
+  zero_mem' := by aesop
+  smul_mem' := by aesop
+
+namespace InverseLimit
+
+@[simp]
+lemma component_eq_component_by_func {g : InverseLimit obj func} {i j : ι} {h : i ≤ j}:
```
YaelDillies, line 47:
```suggestion
lemma func_component {g : InverseLimit obj func} {i j : ι} {h : i ≤ j}:
```

FLT/Deformations/Algebra/InverseLimit.lean
```diff
@@ -0,0 +1,218 @@
...
+
+namespace Module
+
+variable {R : Type*} [Ring R]
+variable [addCommGroupObj : ∀ {i : ι}, AddCommGroup (obj i)] [∀ i, Module R (obj i)]
+variable (func : ∀ {i j : ι}, i ≤ j → obj j →ₗ[R] obj i)
+
+variable (obj) in
+/-- The inverse limit of an inverse system is the modules glued together along the maps. -/
+def InverseLimit : Submodule R (Π i : ι, obj i) where
```
kbuzzard, line 37: Are you sure you want it as a `Submodule` and not as a type by itself? See this comment from `Mathlib.NumberTheory.NumberField.Basic`: 
```lean
/-- The ring of integers (or number ring) corresponding to a number field
is the integral closure of ℤ in the number field.

This is defined as its own type, rather than a `Subalgebra`, for performance reasons:
looking for instances of the form `SMul (RingOfIntegers _) (RingOfIntegers _)` makes
much more effective use of the discrimination tree than instances of the form
`SMul (Subtype _) (Subtype _)`.
The drawback is we have to copy over instances manually.
-/
def RingOfIntegers : Type _ :=
  integralClosure ℤ K
```
Without `Type _` it would have type ` Subalgebra \Z K` or something. The problem with leaving it as a subobject is that 9 times out of 10 with rings of integers, you want to treat them as a type not a term, so you write `a : 𝓞 K` and if you've defined it as a subalgebra then it comes out as `weird up-arrow 𝓞 K`, and whenever you write something like `a • x` typeclass inference is quick to think "hmm, a is in a subtype so if K acted on x then I could use this action" and we would be forever wasting out time looking for actions of K on things like fractional ideals, which don't exist. 

I think basically I'm saying "are you using this as a submodule 9 times out of 10, or as a type? If a type, then you might want to consider
```
def InverseLimit : Type _ := (... : Submodule R (Π i : ι, obj i))
```

FLT/Deformations/Algebra/InverseLimit.lean
```diff
@@ -0,0 +1,218 @@
...
+  unfold InverseLimit at g
+  aesop
+
+instance inhabited : Inhabited (InverseLimit obj func) :=
+  ⟨0⟩
+
+instance unique [IsEmpty ι] : Unique (InverseLimit obj func) where
+  default := ⟨default, by
+    unfold InverseLimit
+    simp
```
kbuzzard, line 58: Does `simp [InverseLimit]` work?

FLT/Deformations/Algebra/InverseLimit.lean
```diff
@@ -0,0 +1,218 @@
...
+end Module
+
+namespace Ring
+
+variable [∀ i : ι, Ring (obj i)]
+variable (func : ∀ {i j}, i ≤ j → obj j →+* obj i)
+
+variable (obj) in
+/-- The inverse limit of an inverse system is the rings glued together along the maps. -/
+def InverseLimit : Subring (Π i : ι, obj i) where
```
kbuzzard, line 93: Similarly, why not consider making this a type?

FLT/Deformations/Algebra/InverseLimit.lean
javierlcontreras: Oh I was unaware of this, but makes sense. I think making it a type makes more sense. Thank you for the detailed explanation :)

kbuzzard:
I'm happy to build "R theory" via `Representation`, at least for now. There are so many ways to do representation theory but I think that this way might be convenient for deformation theory. Let's see! Thanks a lot Javier!

PR #340, opened 2025-02-09
Adds CommAlgebraCat, cloning the Mathlib AlgebraCat file and replacing Ring by CommRing

javierlcontreras:
This is a clone of a subset of Mathlib.Algebra.Category.AlgebraCat.Basic file but with Ring replaced by CommRing. Given the amount of code duplication, this is not expected to be Mathlib-grade code. A better design choice would be to use CommRingCat.Under or FullSubcategory of AlgebraCat, but those APIs are too hard to use right now (for me)

### DIFF
--- a/FLT/Deformations/Algebra/Category/AlgebraCat/CommAlgebraCat.lean
+++ b/FLT/Deformations/Algebra/Category/AlgebraCat/CommAlgebraCat.lean
@@ -8,7 +8,7 @@ import Mathlib.Algebra.FreeAlgebra
 import Mathlib.Algebra.Category.Ring.Basic
 import Mathlib.Algebra.Category.ModuleCat.Basic
 import Mathlib.RingTheory.Ideal.Defs
-import FLT.Mathlib.Algebra.Category.Ring.Basic
+import FLT.Mathlib.Algebra.Algebra.Defs
 
 /-!
 # Category instance for algebras over a commutative ring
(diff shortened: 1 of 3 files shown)


FLT/Deformations/Algebra/Category/AlgebraCat/CommAlgebraCat.lean
```diff
@@ -0,0 +1,231 @@
+/-
+Copyright (c) 2020 Kim Morrison. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Kim Morrison
```
kbuzzard, line 4: This should be you and 2025!

kbuzzard (commented):
This one is difficult. I have been encouraging Ivan Farabella (who is also representing functors on commutative R-algebras) to use `Under`. What exactly is wrong with using `Under`? Andrew has got the hang of it and so you can always bug him if you get stuck :-) I seem to have got on top of it recently. The main problem with it is that if `A : Under R` then IIRC you don't get `[Algebra R A]` for free, but your PR will not solve this either.

javierlcontreras:
In the end, I don't think I will need to use this code. In the end, I have implemented the "Domain Category" of the Deformation Functor (a.k.a C := Residual, Proartinian, Commutative topological O-algebras) from scratch. 

Before this, I was creating it as the WideSubcategory (morphisms are continuous) of the FullSubcategory (Residual, Proartinian) of  CommAlgebraCat. This increased the complexity a lot, due to missing API.

The new code will be upstreamed in the next few days, we can comment in that PR.

PR #343, opened 2025-02-13
chore: fix caching for the docs

YaelDillies:
I forgot to update the cache folder in https://github.com/ImperialCollegeLondon/FLT/pull/334.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -67,6 +67,7 @@ jobs:
 
       - name: Get Mathlib cache
         run: ~/.elan/bin/lake exe cache get || true
+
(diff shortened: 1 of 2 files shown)


pitmonticone (approved):

.github/workflows/blueprint.yml
```diff
@@ -67,14 +67,13 @@ jobs:
 
       - name: Get Mathlib cache
         run: ~/.elan/bin/lake exe cache get || true
```
YaelDillies, line 69:
```suggestion
        run: ~/.elan/bin/lake exe cache get || true

```

pitmonticone (approved):

PR #345, opened 2025-02-14
chore: change from FLT/FLT_files.lean to more traditional FLT.lean

kbuzzard:
@YaelDillies @pitmonticone this is an attempt to make the project align with current standards of what a project should look like. Did I miss anything? It's just moving files around and fixing the `mk_all` script.


pitmonticone:
`import FLT.Deformations.Algebra.InverseLimit` and `import FLT.NumberField.WeakApproximation` are missing in FLT.lean. 

Is it voluntary?

pitmonticone:
Second question, is the script `scripts/mk_all.sh` still necessary? 

We can just run `lake exe mk_all`, right?

kbuzzard:
Re the missing files -- I knew about this, I just wanted to do the move first before I ran the script.

Re using lake exe mk_all -- oh maybe!

Right now the problem I'm running into is that CI didn't like

```
[[lean_lib]]
name = "FLT"
globs = ["FermatsLastTheorem"]
```
and I don't know if it will like the fix I've tried.

pitmonticone:
Yes, adding `"FermatsLastTheorem"` to the `globs` should fix your previous issue.

YaelDillies:
Great!

PR #347, opened 2025-02-14
InverseLimit topology

javierlcontreras:
Gives TopologicalSpace structure to (Module, Ring, Group).InverseLimit. 

Adds TopologicalModule predicate, following the pattern of the Mathlib objects TopologicalGroup and TopologicalRing

Gives TopologicalModule, TopologicalRing and TopologicalGroup structures to each of the  (Module, Ring, Group).InverseLimit respectively


FLT/Deformations/ContinuousRepresentation/TopologicalModule.lean
```diff
@@ -0,0 +1,6 @@
+import Mathlib.Topology.Algebra.Ring.Basic
+
+class TopologicalModule
+  (R : Type*) [Ring R] [TopologicalSpace R] [TopologicalRing R]
+  (M : Type*) [AddCommGroup M] [Module R M] [TopologicalSpace M] where
+  smul_continuous : Continuous (X := R × M) (Y := M) fun ⟨r, m⟩ ↦ r • m
```
kbuzzard, line 6: Indent 4 spaces not 2 on lines 4,5. And `TopologicalModule` should also have `ContinuousAdd M` mathematically. Your `smul_continuous` is just `ContinuousSMul R M` so probably you don't need a new class for this, or you could just extend `ContinuousAdd` and `ContinuousSMul`?

javierlcontreras: Fixed!

javierlcontreras:
The CI is giving build error, but I can't reproduce it locally. Does anyone understand what's the issue?

kbuzzard:
Try merging `main` and see if the problem goes away?

FLT/Deformations/Algebra/InverseLimit/Basic.lean
```diff
@@ -49,10 +50,10 @@ instance instCoeOutPi : CoeOut (InverseLimit obj func) ((i : ι) → obj i) wher
 
 variable {func} in
 @[simp]
-lemma definingProp {a : InverseLimit obj func} : ∀ (i j : _) (h : i ≤ j), func h (a.val j) = a.val i := a.prop
+lemma definingProp {a : InverseLimit obj func} : ∀ {i j} (h : i ≤ j), func h (a.val j) = a.val i := a.prop
```
kbuzzard, line 53: `definingProp` is longer than `prop` and the proof is just `a.prop`. Why make this lemma?

javierlcontreras: The (bad) reason was that I could not remember what the syntax was to add a.prop to simp

FLT/Mathlib/Order/Defs/Unbundled.lean
```diff
@@ -0,0 +1,15 @@
+import Mathlib.Order.Defs.Unbundled
+import Mathlib.Order.CompletePartialOrder
+
+variable {ι : Type*} [instPreorder : Preorder ι]
+
+variable [instDedidabeRel : DecidableRel LE.le (α := ι)] [instIsLinearOrder : IsLinearOrder ι LE.le]
+
+instance : LinearOrder ι where
```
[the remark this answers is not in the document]
javierlcontreras: I thought it was weird that it didn't already exist. I don't fully understand the design choice in this corner of mathlib. I thought the "good" pattern was "first build the house, then paint it". But having `variable [Preorder \i] [LinearOrder \i]` does not promote the existing `LE.le` binary relation, but creates a new one that shadows the old one, right? 

As to why I needed: it is not essential at all, it was just convenient because it let me keep the instance `instBasis` (where I need iota upgraded to a Linear Order) in the same file. I can just move it to a separate file.

FLT/Deformations/Algebra/InverseLimit/Topology.lean
```diff
@@ -0,0 +1,343 @@
...
+  exact (maps_cont i).isOpen_preimage (Exists.choose_spec hV).choose hRo
+
+lemma minimumOpens_isOpen {s : Set (InverseLimit obj func)} : s ∈ minimumOpens func → IsOpen s :=
+  GenerateOpen.basic s
+
+@[continuity]
+lemma toComponent_continuous (i : ι) : Continuous (toComponent func i) where
+  isOpen_preimage R hR := by
+    refine minimumOpens_isOpen func ?_
+    unfold minimumOpens
```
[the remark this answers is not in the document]
javierlcontreras: Fixed!

FLT/Deformations/Algebra/InverseLimit/Topology.lean
```diff
@@ -0,0 +1,343 @@
...
+
+namespace Module.InverseLimit
+
+variable {R : Type*} [Ring R] [∀ i : ι, AddCommGroup (obj i)] [∀ i : ι, Module R (obj i)]
+  [∀ i : ι, TopologicalSpace (obj i)]
+  (func : ∀ {i j}, i ≤ j → obj j →ₗ[R] obj i)
+  (cont : {i j : ι} → (h : i ≤ j) → Continuous (func h))
+
+def minimumOpens : Set (Set (InverseLimit obj func)) :=
+    setOf fun V ↦ ∃ (i : ι) (W : Set (obj i)), IsOpen W ∧ V = (toComponent func i) ⁻¹' W
```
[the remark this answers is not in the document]
javierlcontreras: Fixed!

FLT/Deformations/Algebra/InverseLimit/Basic.lean
```diff
@@ -96,21 +97,51 @@ instance unique [IsEmpty ι] : Unique (InverseLimit obj func) where
...
+lemma func_toComponent {i j : ι} {h : i ≤ j}:
+    .comp (func h) (toComponent func j) = toComponent func i := by
+  unfold toComponent
+  aesop
+
+variable {X : Type*}
+
+def map_of_maps (maps : (i : ι) → X → obj i)
+    (comm : ∀ {i j} (h : i ≤ j) x, (func h) ((maps j) x) = (maps i) x)
+    : X → InverseLimit obj func := fun x ↦ ⟨fun i ↦ maps i x, by aesop⟩
```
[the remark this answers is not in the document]
javierlcontreras: There are many many instances of this in my code. I'll open a new Fix PR changing all of them together. 

I thought the correct indentation was
```
lemma lemma_name (arg1 : type1) 
  (arg2 : type2) (arg3 : type3) (arg4: type4)
  (arg5 : type5) (arg6 : type6) (arg7: type7)
    : type :=
  body
```

But it actually is
```
lemma lemma_name (arg1 : type1) 
  (arg2 : type2) (arg3 : type3) (arg4: type4)
  (arg5 : type5) (arg6 : type6) (arg7: type7) :
    type :=
  body
```

Is this correct? 

I've also seen instances of
```
lemma lemma_name (arg1 : type1) 
    (arg2 : type2) (arg3 : type3) (arg4: type4)
    (arg5 : type5) (arg6 : type6) (arg7: type7) :
  type :=
    body
```

FLT/Deformations/Algebra/InverseLimit/Basic.lean
javierlcontreras: I think I've found it (checking now) `attribute [simp] InverseLimit.prop`

FLT/Deformations/Algebra/InverseLimit/Topology.lean
```diff
@@ -0,0 +1,286 @@
...
+
+variable {R : Type*} [Ring R] [∀ i : ι, AddCommGroup (obj i)] [∀ i : ι, Module R (obj i)]
+  [∀ i : ι, TopologicalSpace (obj i)]
+  (func : ∀ {i j}, i ≤ j → obj j →ₗ[R] obj i)
+  (cont : {i j : ι} → (h : i ≤ j) → Continuous (func h))
+
+abbrev minimumOpens : Set (Set (InverseLimit obj func)) :=
+    {V | ∃ (i : ι) (W : Set (obj i)), IsOpen W ∧ V = (toComponent func i) ⁻¹' W}
+
+instance : TopologicalSpace (InverseLimit obj func) := .generateFrom <| minimumOpens func
```
erdOne, line 21: I think the subspace (of product) topology is the right one? If so, I think that will be easier to use.

FLT/Deformations/Algebra/InverseLimit/Basic.lean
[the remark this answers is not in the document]
kbuzzard: Colons at the end of a line are to be avoided unless it pushes things over 100 chars to put the type after the colon. The indentation is "always 4 spaces before the colon, and two spaces after", so
```lean
lemma lemma_name (arg1 : type1) 
    (arg2 : type2) (arg3 : type3) (arg4: type4)
    (arg5 : type5) (arg6 : type6) (arg7: type7) :
    type :=
  body
```
if you really have to end a line with a colon. I am a bit confused about why I can't find this in the mathlib style guide.

FLT/Deformations/Algebra/InverseLimit/Topology.lean
javierlcontreras: I considered this, but I thought this one would be easier to use, because it is more "categorical". Its by definition the coarsest topology that makes the component maps continuous, making the construction valid in the category of Topological [Rings, Groups, Modules] etc. 

It coincides with the subspace topology of the product for Rings, Groups, Modules but it will not coincide in geeneral (it will not be a subobject of the product in general). Maybe its worth having both?

javierlcontreras: An example where having the "categorical" topology was nice: in the proofs of InverseLimit.instTopologicalModule, InverseLimit.instTopologicalRing, InverseLimit.instTopologicalGroup. They were all very easy diagram chases because you had the nice subbasis `minimumOpens`

kbuzzard: I suspect that the definition of the subspace of the product topology, when unravelled, is also the coarsest topology that makes the component maps continuous. Furthermore there will be a bunch of API for it.

FLT/Deformations/Algebra/InverseLimit/Basic.lean
kbuzzard: I'm surprised this doesn't work. Can you show an example where you need `definingProp`?

FLT/Deformations/Algebra/InverseLimit/Topology.lean
```diff
@@ -0,0 +1,286 @@
...
+  let i := hV.choose
+  let R := hV.choose_spec.choose
+  have hRo := hV.choose_spec.choose_spec.1
+  have hR := hV.choose_spec.choose_spec.2
+  rw [hR]
+  have hcomp : (toComponent func i) ∘ (map_of_maps func maps comm) = maps i := by
+    ext x
+    simp
+  rw [← Set.preimage_comp, hcomp]
+  exact (maps_cont i).isOpen_preimage (Exists.choose_spec hV).choose hRo
```
kbuzzard, line 38:
```suggestion
instance : TopologicalSpace (InverseLimit obj func) := by
  unfold InverseLimit
  infer_instance

@[continuity]
lemma map_of_maps_continuous {X : Type*} [TopologicalSpace X]
    (maps : (i : ι) → X → obj i) (comm : _) (maps_cont : (i : ι) → Continuous (maps i)) :
    Continuous (map_of_maps func maps comm) := by
  unfold InverseLimit
  unfold map_of_maps
  fun_prop
```

Your definitions are clogging up the system and all the proofs are already there. I think you should not make this new definition of a topology, there is nothing wrong with the current definitions in the system. Furthermore, if you ever unfold anything then there is a risk of diamonds if you start putting in new definitions.

FLT/Deformations/Algebra/InverseLimit/Basic.lean
javierlcontreras: I never use it explicitly, I just wanted to add the lemma to `@[simp]` and I don't know more succint syntax.

FLT/Deformations/Algebra/InverseLimit/Topology.lean
javierlcontreras: Big refactor solves this

FLT/Deformations/Algebra/InverseLimit/Topology.lean
javierlcontreras: I don't fully understand what you mean by "all the proofs are already there". I refactored the code so that the topology is defined as the subspace of the product, but I haven't been able to simplify the rest of the proofs (the way I prove them now is by first proving that the subspace topology of the product is the same as the "minimumOpens" topology, and then using the old proofs)

javierlcontreras: I think you might mean using `Subgroup.instTopologicalGroupSubtypeMem` but I have not found a way to do this, because in the definition of InverseLimit, I coerced into a Type for efficiency issues with subobjects. I'm not sure how to use Lemmas like the above when I have purposely forgotten InverseLimit is a subobject

javierlcontreras:
Closing this, everything was reimplemented in a smarter way in https://github.com/ImperialCollegeLondon/FLT/pull/428

PR #349, opened 2025-02-15
Issue #341: Refactor QuotientAddGroup.continuousAddEquiv

### LINKED ISSUE #341: Refactor QuotientAddGroup.continuousAddEquiv
This def in `FLT/Mathlib/Topology/Algebra/Group/Quotient.lean` should be refactored so that (a) it is a lemma about multiplicative groups (and the original def is obtained using `@[to_additive]`) and (b) it uses `comap` instead of `map` i.e. `AddSubgroup.comap e H' = G'`.

smmercuri:
Closes #341


kbuzzard:
Thanks!

PR #353, opened 2025-02-24
Prove prodAdicCompletionComap_isModuleTopology

### LINKED ISSUE #327: Product of L_w has the K_v-module topology.
Say L/K is a finite extension of number fields and v is a finite place of K. I claim that prod_{w|v}L_w has the K_v-module topology. This is `IsDedekindDomain.HeightOneSpectrum.prodAdicCompletionComap_isModuleTopology` in the file `FLT.DedekindDomain.FiniteAdeleRing.BaseChange`. The proof is to firstly observe that the product is finite, and then to invoke the theorem that a finite product of modules with the module topology has the module topology (which we have in mathlib). The fact that L_w has the module topology is #326 so can be assumed, making the work here simply the claim that the number of w's over a given v is finite. This is some presumably standard result about Dedekind domains.

Ruben-VandeVelde:
Fixes #327.

### DIFF
--- a/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
+++ b/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
@@ -246,23 +246,29 @@ lemma v_adicCompletionComapSemialgHom
   subst hvw
   rw [← valuation_comap A K L B w a]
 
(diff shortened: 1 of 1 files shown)


kbuzzard:
Thanks! Will merge after the mathlib bump.

PR #359, opened 2025-02-27
[CI]: fix docs generation

pitmonticone:


scripts/build_docs.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
```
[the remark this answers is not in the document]
pitmonticone: Sure, thanks.

scripts/build_docs.sh
```diff
@@ -29,22 +35,16 @@ template > docbuild/lakefile.toml
...
-mv docs/docs docbuild/.lake/build/doc
-
 # Initialise docbuild as a Lean project
 cd docbuild
-MATHLIB_NO_CACHE_ON_UPDATE=1 # Disable an error message due to a non-blocking bug. See Zulip
-~/.elan/bin/lake update FLT
-~/.elan/bin/lake exe cache get
+
+# Disable an error message due to a non-blocking bug. See Zulip
+MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update FLT
```
kbuzzard, line 42: Just echoing Ruben's comment on Zulip -- is there a reason we're not getting cache any more?

scripts/build_docs.sh
```diff
@@ -29,22 +35,16 @@ template > docbuild/lakefile.toml
...
 
-# Move them out of docbuild
+# Copy documentation to `docs/docs`
 cd ../
-mv docbuild/.lake/build/doc docs/docs
-
-# Clean up after ourselves
-rm -rf docbuild
+sudo chown -R runner docs
+cp -r docbuild/.lake/build/doc docs/docs
```
kbuzzard, line 50: So we are no longer cleaning up after ourselves?

scripts/build_docs.sh
pitmonticone: Yes, because it's unnecessary since `lake update` already does that.

scripts/build_docs.sh
pitmonticone: Again, unnecessary since we cache `docbuild` and we don't want to remove it.

PR #360, opened 2025-02-27
fix: TotallyDefiniteQuaternionAlgebra.AutomorphicForm.add should not be an instance

Ruben-VandeVelde:
Noticed because it causes an error in documentation generation:

PANIC at _private.DocGen4.Process.InstanceInfo.0.DocGen4.Process.InstanceInfo.ofDefinitionInfo DocGen4.Process.InstanceInfo:50:39: isClass? on TotallyDefiniteQuaternionAlgebra.AutomorphicForm.add returned none


kbuzzard:
huh! Thanks!

PR #362, opened 2025-02-27
K is discrete in the adeles of K (#257)

### LINKED ISSUE #257: If K is a number field then K is discrete in the adeles of K
Assuming discreteness for K=Q (an earlier result), and the fact that A_K is K-isomorphic to K tensor A_Q (another earlier result), we can deduce discreteness for all K. Slightly more details are [in the blueprint](https://imperialcollegelondon.github.io/FLT/blueprint/Adele_miniproject.html#NumberField.AdeleRing.discrete).

matthewjasper:
This swaps the theorem order in the blueprint around to generalize to all number fields and then to all elements, hopefully that's acceptable.

Closes #257

### DIFF
--- a/FLT/NumberField/AdeleRing.lean
+++ b/FLT/NumberField/AdeleRing.lean
@@ -28,110 +28,6 @@ instance NumberField.AdeleRing.locallyCompactSpace : LocallyCompactSpace (AdeleR
 
 end LocallyCompact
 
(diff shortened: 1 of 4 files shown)


kbuzzard:
Yeah this way is fine. Thanks for updating the blueprint too!

PR #365, opened 2025-03-01
Drop LinearMap.det_mul

Ruben-VandeVelde:
@eric-wieser suggested not to upstream this, and it seems unnecessary indeed.


eric-wieser:
(in https://github.com/leanprover-community/mathlib4/pull/22431)

PR #368, opened 2025-03-04
Unpin lean version from `lakefile.toml`

smmercuri:


pitmonticone (approved):

kbuzzard:
I don't really know what I'm talking about here but do either of `lean-toolchain` and `lake-manifest.json` also need to be changed? They were both changed in the bump PR.

PR #369, opened 2025-03-05
chore: bump toolchain to v4.18.0-rc1

pitmonticone:


lake-manifest.json
```diff
@@ -15,17 +15,17 @@
    "type": "git",
    "subDir": null,
    "scope": "",
-   "rev": "5269898d6a51d047931107c8d72d934d8d5d3753",
+   "rev": "26aae2611fd03b7034664913a9f63e55333970a9",
    "name": "mathlib",
    "manifestFile": "lake-manifest.json",
-   "inputRev": "v4.17.0",
+   "inputRev": null,
```
kbuzzard, line 21: Is this really right?

lakefile.toml
```diff
@@ -9,7 +9,6 @@ relaxedAutoImplicit = false # switch off relaxed auto-implicit
 [[require]]
 name = "mathlib"
 git = "https://github.com/leanprover-community/mathlib4.git"
-rev = "master"
```
kbuzzard, line 12: Is this really right?

pitmonticone: Yes, the `rev` field it's unnecessary in our case. When we want to pin the `stable` version we can always add it.

PR #384, opened 2025-03-29
L tensor K_v = \oplus_{w|v} L_w is a homeomorphism (#328)

### LINKED ISSUE #328: L tensor K_v = \oplus_{w|v} L_w is a homeomorphism
This is an easy one mathematically, because all the ingredients are there. Both sides are K_v-algebras; the left hand side has the K_v-module topology by definition, the right hand side has a K_v-algebra structure coming from the map K_v -> \prod_{w|v} L_w which this map came from, and we've shown in earlier results that the topology on the sum is the K_v-module topology. Now the result follows from the fact that K_v-linear maps between modules with the K_v-module-topology are automatically continuous.

matthewjasper:
* Add proofs for adicCompletionComapContinuousAlgEquiv
* Add ContinuousSMul instance for `\oplus_{w|v} L_w`

closes #328

### DIFF
--- a/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
+++ b/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
@@ -425,12 +425,39 @@ variable (v : HeightOneSpectrum A) in
 instance : IsModuleTopology (adicCompletion K v) (L ⊗[K] adicCompletion K v) :=
   ⟨rfl⟩
 
-noncomputable def adicCompletionComapContinuousAlgEquiv (v : HeightOneSpectrum A) :
-  L ⊗[K] v.adicCompletion K ≃A[L] ∀ w : v.Extension B, w.1.adicCompletion L
+attribute [local instance] Algebra.TensorProduct.rightAlgebra in
+/-- `adicCompletionComapAlgEquiv` as a `K_v`-algebra equivalence -/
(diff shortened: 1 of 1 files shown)


FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -376,6 +376,15 @@ noncomputable instance comap_pi_algebra (v : HeightOneSpectrum A) :
     Algebra (v.adicCompletion K) (Π (w : v.Extension B), w.1.adicCompletion L) :=
   (adicCompletionComapSemialgHom' A K L B v).toAlgebra
 
+attribute [local instance] comap_algebra comap_algebra_continuousSmul in
+instance comap_pi_continuousSmul (v : HeightOneSpectrum A) :
+    letI := comap_pi_algebra A K L B v |>.toSMul
+    ContinuousSMul (adicCompletion K v) (Π (w : v.Extension B), w.1.adicCompletion L) := by
```
matthewjasper, line 382: We're not very consistent on how to write `Π (w : v.Extension B), w.1.adicCompletion L`.  I've tried to match whatever the closest existing use had, but it's probably worth making them all match.

kbuzzard:
Oh nice! We are really making progress here! Thanks a lot!

PR #389, opened 2025-04-01
chore: remove "import Mathlib"

kbuzzard:
We're spending too long building irrelevant docs. This will speed things up.


kbuzzard:
Indeed the docs now build twice as quickly.

PR #391, opened 2025-04-01
chore: update lakefile.toml to compile FLT file, not FLT dir

kbuzzard:
Now we use FLT.lean we don't need to compile `FLT.*`, we can just compile `FLT`.

### DIFF
--- a/lakefile.toml
+++ b/lakefile.toml
@@ -16,4 +16,4 @@ git = "https://github.com/PatrickMassot/checkdecls.git"
 
 [[lean_lib]]
 name = "FLT"
-globs = ["FLT.*", "FermatsLastTheorem"]
+globs = ["FLT", "FermatsLastTheorem"]


digama0:
One danger of dropping the glob is that you can end up with files that don't compile under `FLT/Foo.lean` as long as they are not imported by `FLT.lean`. So I would recommend either reusing mathlib's `mk_all` script or keeping the glob.

kbuzzard:
Yes, we have `mk_all` working here. But do you mean we should be checking it in CI? I'm not sure we're doing that.

kbuzzard:
Before we merge this we should probably get CI to check that `mk_all` doesn't change `FLT.lean`.

kbuzzard:
This is waiting on #427 which should ensure that CI fails on PRs which don't have FLT.lean which corresponds exactly to the files in the FLT directory.

kbuzzard:
OK I have now seen PRs failing with my own eyes with the wrong FLT.lean so I think this is safe to go.

PR #398, opened 2025-04-05
Bump mathlib to v4.19.0-rc2

YaelDillies:
You can check this is bumped to the `v4.19.0-rc2` tag of mathlib because of the `inputRev` change in `lake-manifest.json`

Co-authored-by: Pietro Monticone


pitmonticone (approved):

kbuzzard:
I've usually not been too fussed about whether we bump to a specific tag or just to some random mathlib commit, but for the first time today I discovered I did care: I have a PR to mathlib refactoring finite adeles and when I wanted to see the fallout for this repo I discovered that there were a bunch of unrelated errors because the version of mathlib which I'd branched off was not exactly the same as the version of mathlib which we'd bumped to (in fact FLT was ahead!). So I can see the advantage of both "random bumps" (to keep up) and "precise tag bumps". I'll bump the mathlib adele refactor branch to precisely this tag too which will solve my adele refactor problems. Thanks to both of you!

YaelDillies:
To be clear, it's okay not being fussed. But then you shouldn't pretend that you are fussed by creating `v4.something` tags. Those should only be used when the versions precisely match.

kbuzzard:
ok got it!

Issue #329, opened 2025-02-01
The map B tensor_A O_v -> \prod_{w|v} O_w is an isomorphism for almost all v


kbuzzard:
The proof is in Cassels-Froehlich and is described in the blueprint (or will be when I push to main), in label `  \label{IsDedekindDomain.HeightOneSpectrum.adicCompletionComapAlgEquiv_integral}`. The sorry is `IsDedekindDomain.HeightOneSpectrum.adicCompletionComapAlgEquiv_integral` in `FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean`.

matthewjasper:
claim

kbuzzard:
propose #400

PR #400, opened 2025-04-06
The map B tensor_A O_v -> \prod_{w|v} O_w is an isomorphism (#329)

matthewjasper:
Not the proof in the blueprint to avoid having to work with discriminants and to avoid bases as much as possible.
This also doesn't need the finite set of exceptions.
Still needs some work, but the proofs probably won't change too much now.

### DIFF
--- a/FLT/DedekindDomain/AdicValuation.lean
+++ b/FLT/DedekindDomain/AdicValuation.lean
@@ -8,9 +8,9 @@ import Mathlib.Algebra.Algebra.Subalgebra.Pi
 
 # Adic Completions
 
-If `A` is a valued ring with field of fractions `K` there are two ways one
-might define the integers of the completion `K_v`, first by
-`𝒪_v = {x ∈ K_v | v x ≤ 1}` and second by taking the `v-adic` completion of `A`.
+If `A` is a valued ring with field of fractions `K` there are two different
+complete rings containing `A` one might define, the first is
+`𝒪_v = {x ∈ K_v | v x ≤ 1}` and the second is the `v-adic` completion of `A`.
 In the case when `A` is a Dedekind domain these definitions give isomorphic
 topological `A`-algebras. This file makes some progress towards this.
 
--- a/FLT/DedekindDomain/AdicValuation.lean
+++ b/FLT/DedekindDomain/AdicValuation.lean
@@ -132,7 +132,7 @@ lemma exists_adicValued_sub_lt_of_adicValued_le_one { x : (WithVal (v.valuation 
       HeightOneSpectrum.intValuation_apply]
   apply lt_of_le_of_lt hval hku
 
-/-- The closure of `A` in `K_v` is `O_[K_v]`. -/
+/-- The closure of `A` in `K_v` is `𝒪_v`. -/
 theorem closureAlgebraMapIntegers_eq_integers :
     closure (algebraMap A (v.adicCompletion K)).range =
     SetLike.coe (v.adicCompletionIntegers K) := by
@@ -140,7 +140,7 @@ theorem closureAlgebraMapIntegers_eq_integers :
   . apply closure_minimal _ Valued.valuationSubring_isClosed
     rintro b ⟨a, rfl⟩
     exact HeightOneSpectrum.coe_mem_adicCompletionIntegers v a
-  -- Show `O_[K_v] ⊆ closure A` from `O_[K_v] ⊆ closure O_[K]` and `closure O_[K] ⊆ closure A`
+  -- Show `𝒪_v ⊆ closure A` from `𝒪_v ⊆ closure O_[K]` and `closure O_[K] ⊆ closure A`
   let f := fun (k : WithVal (v.valuation K)) => (k : v.adicCompletion K)
   suffices h : closure (f '' (f ⁻¹' (HeightOneSpectrum.adicCompletionIntegers K v))) ⊆
       closure (algebraMap A (HeightOneSpectrum.adicCompletion K v)).range by
@@ -165,7 +165,7 @@ theorem closureAlgebraMapIntegers_eq_integers :
   apply hγ
   simpa
 
-/-- A is dense in `O_[K_v]`. -/
+/-- A is dense in `𝒪_v`. -/
 theorem denseRange_of_integerAlgebraMap :
     DenseRange (algebraMap A (v.adicCompletionIntegers K)) := by
   rw [denseRange_iff_closure_range]
@@ -179,7 +179,7 @@ theorem denseRange_of_integerAlgebraMap :
   simp only [RingHom.coe_range, ← Set.range_comp']
   rfl
 
-/-- An element of `O_[K_v]` can be approximated by an element of `A`. -/
+/-- An element of `𝒪_v` can be approximated by an element of `A`. -/
 theorem exists_adicValued_sub_lt_of_adicCompletionInteger ( x : v.adicCompletionIntegers K )
     ( γ : (WithZero (Multiplicative ℤ))ˣ ) :
     ∃a, Valued.v ((algebraMap A K a) - (x : v.adicCompletion K)) < γ.val := by
@@ -196,7 +196,7 @@ theorem exists_adicValued_sub_lt_of_adicCompletionInteger ( x : v.adicCompletion
   rw [HeightOneSpectrum.algebraMap_adicCompletion, Function.comp_apply] at ha
   rwa [ha]
 
-/-- An element of `∏_{v ∈ s} O_[K_v]`, with `s` finite, can be approximated by an element of `A`.
+/-- An element of `∏_{v ∈ s} 𝒪_v`, with `s` finite, can be approximated by an element of `A`.
   -/
 theorem exists_forall_adicValued_sub_lt {ι : Type*} (s : Finset ι)
     (e : ι → (WithZero (Multiplicative ℤ))ˣ ) (valuation : ι → HeightOneSpectrum A)
@@ -253,7 +253,7 @@ theorem exists_forall_adicValued_sub_lt {ι : Type*} (s : Finset ι)
     add_comm_sub, add_sub, eq_sub_iff_add_eq]
   rfl
 
-/-- The closure of `A` in `∏_{v ∈ s} K_v` is `∏_{v ∈ s} O_[K_v]`. `s` may be infinite. -/
+/-- The closure of `A` in `∏_{v ∈ s} K_v` is `∏_{v ∈ s} 𝒪_v`. `s` may be infinite. -/
 theorem closureAlgebraMapIntegers_eq_prodIntegers {ι : Type*}
     (valuation : ι → HeightOneSpectrum A) (injective : Function.Injective valuation) :
     closure (SetLike.coe (algebraMap A ((i : ι) → (valuation i).adicCompletion K)).range) =
--- a/FLT/DedekindDomain/AdicValuation.lean
+++ b/FLT/DedekindDomain/AdicValuation.lean
@@ -34,12 +34,15 @@ lemma emultiplicity_eq_of_valuation_eq_ofAdd { a : A } { k : ℕ } ( hnz : a ≠
   rw [← hv, UniqueFactorizationMonoid.emultiplicity_eq_count_normalizedFactors v.irreducible hnb,
     count_associates_factors_eq hnb v.isPrime v.ne_bot, normalize_eq]
 
-lemma exists_approxInverse { a b : A } { n : ℕ } ( hnz : a ≠ 0 )
+/-- Given `a, b ∈ A` and `v b ≤ v a` we can find `y in A` such that `y` is close to `a / b` by
+    the valuation v. -/
+lemma exists_adicValued_mul_sub_le { a b : A } { n : ℕ } ( hnz : a ≠ 0 )
     ( hle : Multiplicative.ofAdd (-(n : ℤ)) ≤ v.intValuationDef a)
     ( hle' : v.intValuationDef b ≤ v.intValuationDef a ) :
     ∃ y, v.intValuationDef (y * a - b) ≤ (Multiplicative.ofAdd (-(n : ℤ))) := by
   have hnb : Ideal.span {a} ≠ ⊥ := by
     rwa [ne_eq, Ideal.span_singleton_eq_bot]
+  -- Rewrite the statements to involve multiplicity rather than valuations
   rw [intValuation_eq_coe_neg_multiplicity _ hnz, WithZero.coe_le_coe, Multiplicative.ofAdd_le,
     neg_le_neg_iff, Int.ofNat_le] at hle
   have hm : emultiplicity v.asIdeal (Ideal.span {a}) ≤ n := by
@@ -48,15 +51,20 @@ lemma exists_approxInverse { a b : A } { n : ℕ } ( hnz : a ≠ 0 )
   have hb : b ∈ v.asIdeal ^ multiplicity v.asIdeal (Ideal.span {a}) := by
     rwa [← Ideal.dvd_span_singleton, ← HeightOneSpectrum.intValuation_le_pow_iff_dvd,
         ← intValuation_eq_coe_neg_multiplicity _ hnz]
+  -- Now make use of
+  -- `v.asIdeal ^ multiplicity v.asIdeal (Ideal.span {a}) = v.asIdeal ^ n ⊔ Ideal.span {a}`
+  -- (this is where we need IsDedekindDomain A)
   rw [← irreducible_pow_sup_of_ge hnb (HeightOneSpectrum.irreducible v) n hm] at hb
+  -- Extract y by writing b as a general term of the sum of the two ideals.
   obtain ⟨x, hx, z, hz, hxz⟩ := Submodule.mem_sup.mp hb
   obtain ⟨y, hy⟩ := Ideal.mem_span_singleton'.mp hz
   use y
+  -- And again prove the result about valuations by turning into one about ideals.
   rwa [hy, ← hxz, sub_add_cancel_right, HeightOneSpectrum.intValuation_le_pow_iff_dvd,
       Ideal.dvd_span_singleton, neg_mem_iff]
 
 open scoped Multiplicative in
-lemma exists_ofAdd_lt_and_ofAdd_lt { x y : ℤₘ₀ } ( hx : x ≠ 0 ) ( hy : y ≠ 0) :
+private lemma exists_ofAdd_lt_and_ofAdd_lt { x y : ℤₘ₀ } ( hx : x ≠ 0 ) ( hy : y ≠ 0) :
     ∃ (k : ℕ), (Multiplicative.ofAdd (-(k : ℤ))) < x ∧ (Multiplicative.ofAdd (-(k : ℤ))) < y := by
   let u := WithZero.unzero hx
   let v := WithZero.unzero hy
@@ -83,8 +91,10 @@ lemma exists_ofAdd_lt_and_ofAdd_lt { x y : ℤₘ₀ } ( hx : x ≠ 0 ) ( hy : y
 lemma exists_adicValued_sub_lt_of_adicValued_le_one { x : (WithVal (v.valuation K)) }
     ( γ : (WithZero (Multiplicative ℤ))ˣ ) ( hx : Valued.v x ≤ 1 ) :
     ∃a, Valued.v ((algebraMap A K a) - (x : v.adicCompletion K)) < γ.val := by
+  -- Write `x = n / d`
   obtain ⟨⟨n, d, hd⟩, hnd⟩ := IsLocalization.surj (nonZeroDivisors A) x
   dsimp only at hnd
+  -- Show v n ≤ v d
   have hnd' := congr_arg Valued.v hnd
   simp only [HeightOneSpectrum.adicValued_apply', map_mul] at hnd'
   have hge : Valued.v ((algebraMap A (WithVal (v.valuation K))) d) ≥
@@ -99,6 +109,7 @@ lemma exists_adicValued_sub_lt_of_adicValued_le_one { x : (WithVal (v.valuation 
     HeightOneSpectrum.valuation_of_algebraMap] at hge
   have hdz : (algebraMap A (WithVal (v.valuation K)) d) ≠ 0 :=
     IsLocalization.to_map_ne_zero_of_mem_nonZeroDivisors _ (fun _ ↦ id) hd
+  -- Find a suitable `k` for the bound in `exists_adicValued_mul_sub_le`
   have hv : Valued.v ((algebraMap A (WithVal (v.valuation K)) d)) ≠ 0 := by
     rw [Valuation.ne_zero_iff]
     exact hdz
@@ -108,7 +119,8 @@ lemma exists_adicValued_sub_lt_of_adicValued_le_one { x : (WithVal (v.valuation 
   obtain ⟨k, hku, hkv⟩ := exists_ofAdd_lt_and_ofAdd_lt hu hv
   simp only [WithVal, HeightOneSpectrum.adicValued_apply,
     HeightOneSpectrum.valuation_of_algebraMap, HeightOneSpectrum.intValuation_apply] at hkv
-  obtain ⟨a, hval⟩ := exists_approxInverse v (nonZeroDivisors.ne_zero hd) hkv.le hge
+  -- Now can apply `exists_adicValued_mul_sub_le` to get the approximation of x.
+  obtain ⟨a, hval⟩ := exists_adicValued_mul_sub_le v (nonZeroDivisors.ne_zero hd) hkv.le hge
   use a
   rw [← eq_div_iff_mul_eq hdz] at hnd
   rw [← UniformSpace.Completion.coe_sub,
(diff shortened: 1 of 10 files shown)


FLT/Mathlib/Topology/Algebra/Valued/ValuationTopology.lean
```diff
@@ -8,3 +8,16 @@ lemma Valued.isUnit_valuationSubring_iff {Γ₀ : Type*} [LinearOrderedCommGroup
     IsUnit x ↔ Valued.v (x.val : F) = 1 := by
   convert Valuation.Integers.isUnit_iff_valuation_eq_one _
   exact Valuation.integer.integers _
+
+/-- The unit ball of a valued ring is open. -/
```
matthewjasper, line 12:
```suggestion
/-- The unit ball of a valued ring is closed. -/
```

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -493,42 +496,216 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+        rw [Algebra.ofId_apply]
+        have halgMap : (algebraMap A (adicCompletion K v)) a =
+            (algebraMap A K a) • (1 : adicCompletion K v) := by
+          rw [← Algebra.algebraMap_eq_smul_one]
+          rfl
+        rw [halgMap, ← TensorProduct.smul_tmul]
+        simp [← Algebra.smul_def, IsScalarTower.algebraMap_smul]
+
+open TensorProduct.AlgebraTensorModule in
+set_option maxHeartbeats 400000 in
```
matthewjasper, line 571: I need to look at this

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -493,42 +496,198 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+  intro t ⟨s, hs⟩
+  rw [← hs]
+  clear t hs
+  induction s with
+    | zero =>
+        apply subset_closure
+        use 0
+        simp
+    | add x y hx hy =>
+        -- Instances are timing out without help
```
matthewjasper, line 513: This also doesn't work on CI

FLT/DedekindDomain/DenseIntegers.lean
```diff
@@ -0,0 +1,273 @@
...
+  let f := fun (k : WithVal (v.valuation K)) => (k : v.adicCompletion K)
+  have hd : DenseRange f := by
+    apply UniformSpace.Completion.denseRange_coe
+  apply subset_antisymm
+  . apply closure_minimal _ Valued.valuationSubring_isClosed
+    rintro b ⟨a, rfl⟩
+    exact HeightOneSpectrum.coe_mem_adicCompletionIntegers v a
+  suffices h : closure (algebraMap A (HeightOneSpectrum.adicCompletion K v)).range =
+      closure (f '' (f ⁻¹' (HeightOneSpectrum.adicCompletionIntegers K v))) by
+    rw [h]
```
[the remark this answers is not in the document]
matthewjasper: Never mind, that theorem uses less than

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+import Mathlib.Algebra.Algebra.Subalgebra.Pi
+import Mathlib.Algebra.Order.GroupWithZero.Canonical
+
+namespace IsDedekindDomain
+
+variable { A : Type* } ( K : Type* ) [CommRing A] [Field K] [Algebra A K] [IsFractionRing A K]
+    [IsDedekindDomain A] (v : HeightOneSpectrum A)
+
+lemma intValuation_eq_coe_neg_multiplicity { a : A } ( hnz : a ≠ 0 ) :
+    v.intValuationDef a =
```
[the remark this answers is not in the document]
kbuzzard: Yeah good question! `Mathlib/RingTheory/DedekindDomain/AdicValuation.lean` seems to develop API for both, with inconsistent naming as well? But it doesn't seem to be used elsewhere. My instinct would be to use `v.intValuation` everywhere and also make a PR to mathlib making that file more consistent (e.g. either fix the names of lemmas like `intValuation_ne_zero` and 5 or so others which talk about `intValuationDef` not `intValuation`, or just make them about `intValuation` anyway, or maybe even rename `intValuationDef to `intValuation.toFun`.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+    decide
+  apply lt_mul_of_le_of_one_lt
+  . apply le_sup_of_le_right
+    rw [Int.ofNat_toNat, le_sup_iff]
+    left
+    rfl
+  decide
+
+/-- An element of `K` with valuation ≤ 1 can be approximated by an element of `A`.
+  -/
```
kbuzzard, line 90:
```suggestion
-/
```

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  obtain ⟨a, ha⟩ := exists_adicValued_sub_lt_of_adicValued_le_one K v γ hx
+  use algebraMap A K a
+  constructor
+  . use a
+    rfl
+  apply hγ
+  simpa
+
+/-- A is dense in `𝒪_v`. -/
+theorem denseRange_of_integerAlgebraMap :
```
kbuzzard, line 169: Getting this into mathlib would be a very natural project.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+    DenseRange (algebraMap A (v.adicCompletionIntegers K)) := by
+  rw [denseRange_iff_closure_range]
+  ext x
+  simp only [Set.mem_univ, iff_true]
```
kbuzzard, line 173:
```suggestion
  rw [denseRange_iff_closure_range, Set.eq_univ_iff_forall]
  intro x
```

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+      (mul_le_mul_of_nonneg_left bc a0).trans_lt <| mul_lt_of_lt_one_left c0 ha
+    apply mul_lt_of_lt_one_of_le'
+    . norm_cast
+    . rw [← WithZero.coe_inv]
+      apply le_of_le_of_eq _ (WithZero.coe_unzero (e i).ne_zero)
+      rw [WithZero.coe_le_coe, ← ofAdd_neg, neg_sup]
+      apply inf_le_of_left_le
+      apply le_of_eq (Int.neg_neg (WithZero.unzero (e i).ne_zero))
+    . norm_cast
+    apply Units.zero_lt
```
kbuzzard, line 229: You've proved this before, right? This is again just the statement that given an element of Z_m (the nonzero elements of Z_m0), there's an element of -N_m which is less than it. This proof looks even longer than the one you gave the first time, when you proved that given two elements of Z_m there was an element of -N_m which is less than both of them. I think you should just pull out this lemma as a sublemma and just use it both here and in the two-variable case.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  have hinj : ∀ i ∈ s, ∀ j ∈ s, i ≠ j →
+      (fun i ↦ (valuation i).asIdeal) i ≠ (fun i ↦ (valuation i).asIdeal) j := by
+    intro i hi j hj
+    contrapose!
+    intro hij
+    apply injective
+    apply HeightOneSpectrum.ext hij
```
kbuzzard, line 236:
```suggestion
    exact mt <| fun hij ↦ injective (HeightOneSpectrum.ext hij)
```
It's exactly the same but it makes you look much cleverer.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+    (valuation : ι → HeightOneSpectrum A) (injective : Function.Injective valuation) :
+    closure (SetLike.coe (algebraMap A ((i : ι) → (valuation i).adicCompletion K)).range) =
+    (Set.pi Set.univ (fun (i : ι) ↦ HeightOneSpectrum.adicCompletionIntegers K (valuation i))) := by
+  apply Set.Subset.antisymm
+  . apply closure_minimal
+    . rintro c ⟨a, ha⟩ i -
+      rw [← ha]
+      simp only [Pi.algebraMap_apply, SetLike.mem_coe]
+      exact HeightOneSpectrum.coe_mem_adicCompletionIntegers (valuation i) a
+    apply isClosed_set_pi
```
kbuzzard, line 267: Again this argument should be indented with a dot.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+    rintro w -
+    exact Valued.valuationSubring_isClosed
+  intro f hf
```
kbuzzard, line 270: And here.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  rw [mem_closure_iff_nhds_zero]
+  intro U hU
+  rw [Pi.zero_def, nhds_pi, Filter.mem_pi'] at hU
+  obtain ⟨I, t, htn, hts⟩ := hU
+  obtain ⟨g, hg⟩ := Classical.axiomOfChoice
+    (fun w => (Valued.is_topological_valuation (t w)).mp (htn w))
```
kbuzzard, line 276:
```suggestion
  choose g hg using (fun w => (Valued.is_topological_valuation (t w)).mp (htn w))
```
This is the idiomatic way to write what you wrote.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  obtain ⟨a, ha⟩ :=
+    exists_forall_adicValued_sub_lt K I g valuation injective (fun w => ⟨f w, hf w ⟨⟩⟩)
+  use algebraMap A _ a
+  constructor
+  . rw [RingHom.coe_range]
+    exact Set.mem_range_self a
+  apply hts
```
kbuzzard, line 283: . and indent

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  intro w hw
+  apply hg
+  apply ha w hw
+
+end IsDedekindDomain
```
kbuzzard, line 288: Well done! This is a tour de force!

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+        apply subset_closure
+        use algebraMap A B a * b
+        unfold f
+        rw [Algebra.algebraMap_eq_smul_one (A := (adicCompletionIntegers K v)) a,
+          coe_smul_adicCompletionIntegers, ← TensorProduct.smul_tmul, Algebra.ofId_apply,
+          Algebra.TensorProduct.algebraMap_apply, RingHom.map_mul, ← Algebra.smul_def]
+        simp
+
+open TensorProduct.AlgebraTensorModule in
+open scoped Classical in
```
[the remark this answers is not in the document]
kbuzzard: I would discourage `open scoped Classical in` but here it seems like you can just delete it?

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+    [Module.Finite A B] [IsDedekindDomain B] [IsFractionRing B L]  in
+lemma tensorAdicCompletionIsClosedRange :
+    IsClosed (SetLike.coe (tensorAdicCompletionIntegersTo A K L B v).range) := by
+  -- `B ⊗[A] 𝒪_v` is a subgroup of `L ⊗[K] K_v`, so we can show it's closed
+  -- by showing that it's open.
+  rw [← Subalgebra.coe_toSubring, ← Subring.coe_toAddSubgroup]
+  letI : Module (adicCompletion K v) (L ⊗[K] adicCompletion K v) :=
+      Algebra.TensorProduct.rightAlgebra.toModule
+  letI : SMul (adicCompletion K v) (L ⊗[K] adicCompletion K v) :=
+      Algebra.TensorProduct.rightAlgebra.toSMul
```
kbuzzard, line 571: You don't need these, typeclass inference will take care of them because of the local instance on line 560.

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+  use equiv.symm '' (Set.pi Set.univ (fun _ => SetLike.coe (adicCompletionIntegers K v)))
+  refine ⟨?_, ?_, ?_⟩
+  . intro t ⟨g, hg, ht⟩
+    -- We have `t = equiv g = ∑ i, b i ⊗ g i`, since `g in ∏ 𝒪_v` and
+    -- `b i ∈ (algebraMap B L).range`, this is `tensorAdicCompletionIntegersTo`
+    -- of some element of `B ⊗[A] 𝒪_v`
+    have hf : ∀ (i : ι), ∃ (w : B), (algebraMap B L w) = (b i) := by
+      intro i
+      apply IsIntegralClosure.isIntegral_iff.mp (hb i)
+    obtain ⟨f, hf_prop⟩ := Classical.axiomOfChoice hf
```
kbuzzard, line 598:
```suggestion
    choose f hf_prop using hf
```

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+  obtain ⟨ι, b, hb⟩ := FiniteDimensional.exists_is_basis_integral A K L
+  let b' : Basis ι (adicCompletion K v) (L ⊗[K] (adicCompletion K v)) :=
+    Basis.rightBaseChange L b
+  -- Use the basis to get a continuous equivalence from `L ⊗[K] K_v` to `ι → K_v`.
+  let equiv : L ⊗[K] (adicCompletion K v) ≃L[K] (ι → adicCompletion K v) :=
+    IsModuleTopology.continuousLinearEquiv (b'.equivFun) |>.restrictScalars K
+
+  -- Use the preimage of `∏ 𝒪_v` as the open neighbourhood.
+  use equiv.symm '' (Set.pi Set.univ (fun _ => SetLike.coe (adicCompletionIntegers K v)))
+  refine ⟨?_, ?_, ?_⟩
```
kbuzzard, line 590:
```suggestion
  refine ⟨?_, ?_, by simp [ValuationSubring.zero_mem]⟩
```
saves you a line later

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+omit [Algebra.IsSeparable K L] [IsDomain B] [Algebra.IsIntegral A B] [Module.Finite A B]
+    [IsDedekindDomain B] [IsFractionRing B L] in
+lemma tensorAdicCompletionIntegersToRange_eq_closureIntegers :
+    SetLike.coe (tensorAdicCompletionIntegersTo A K L B v).range =
+        closure (algebraMap B (L ⊗[K] adicCompletion K v)).range := by
+  apply Set.Subset.antisymm
+  . apply tensorAdicCompletionIntegersToRange_subset_closureIntegers
+  apply closure_minimal
+  . intro c ⟨b, hb⟩
+    rw [← hb]
```
kbuzzard, line 634:
```suggestion
  . rintro c ⟨b, rfl⟩
```
and again this proof should be more indented.

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -491,42 +495,176 @@ noncomputable def tensorAdicCompletionIntegersTo (v : HeightOneSpectrum A) :
...
+  rw [Subalgebra.coe_pi]
+  show (SetLike.coe (Submodule.pi _  _)) = _
+  rw [Submodule.coe_pi]
+  let val := (fun (w : Extension B v) ↦ w.1)
+  have hinj : Function.Injective val :=
+    (Set.injective_codRestrict Subtype.property).mp fun _ _ a ↦ a
+  rw [closureAlgebraMapIntegers_eq_prodIntegers L val hinj]
+  rfl
+
+set_option synthInstance.maxHeartbeats 40000 in
```
kbuzzard, line 652: Do you know why this is happening? I can take a look later.

FLT/DedekindDomain/AdicValuation.lean
kbuzzard, line 229: You've proved this before, right? This is again just the statement that given an element of Z_m (the nonzero elements of Z_m0), there's an element of -N_m which is less than it. This proof looks even longer than the one you gave the first time, when you proved that given two elements of Z_m there was an element of -N_m which is less than both of them. I think you should just pull out this lemma as a sublemma and just use it both here and in the two-variable case.

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
kbuzzard, line 571: You don't need these, typeclass inference will take care of them because of the local instance on line 560.

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
kbuzzard: Adding
```lean
instance : MulActionHomClass
    (L ⊗[K] adicCompletion K v →ₐ[L] (w : Extension B v) → adicCompletion L w.1) B
    (L ⊗[K] adicCompletion K v) ((w : Extension B v) → adicCompletion L w.1) where
  map_smulₛₗ φ b x := by
    rw [← IsScalarTower.algebraMap_smul L, AlgHom.map_smul_of_tower,
      IsScalarTower.algebraMap_smul, id_def]
```
before the lemma fixes it. I diagnosed this by looking at the instance trace; Lean was spending 1.5 seconds trying and failing to solve this using typeclass inference.

FLT/Mathlib/RingTheory/TensorProduct/Basis.lean
```diff
@@ -0,0 +1,51 @@
...
+variable {R : Type*} (A : Type*) {B : Type*} {ι : Type*} [CommSemiring R]
+variable [CommSemiring A] [Algebra R A] [Fintype ι]
+variable [CommSemiring B] [Algebra R B]
+
+/-- The lift of an `R`-basis of `A` to a `B`-basis of the base change `A ⊗[R] B`. -/
+
+noncomputable
+def Basis.rightBaseChange (b : Basis ι R A) : Basis ι B (A ⊗[R] B) where
+  repr :=
+    letI : DecidableEq ι := Classical.typeDecidableEq ι
```
kbuzzard, line 21: I am not sure that this is right. There could be a competing decidable equality instance on \io if it is for example `Fin n`. I would ask on the Zulip if I were you (or I can ask if you'd rather). You could always put `[DecidableEq \io]` as a hypothesis, this is probably preferable.

FLT/Mathlib/RingTheory/TensorProduct/Basis.lean
```diff
@@ -0,0 +1,51 @@
...
+        (LinearEquiv.refl B B)
+        b.equivFun).trans
+      (TensorProduct.piScalarRight _ _ _ _)
+    let finite : (ι →₀ B) ≃ₗ[B] (ι → B) := Finsupp.linearEquivFunOnFinite B B ι
+    comm.symm.trans π |>.trans finite.symm
+
+@[simp]
+lemma Basis.rightBaseChange_repr (b : Basis ι R A) (i) (x : B) :
+    (b.rightBaseChange A).repr (b i ⊗ₜ x) = Finsupp.single i x := by
+  letI : DecidableEq ι := Classical.typeDecidableEq ι
```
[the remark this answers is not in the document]
kbuzzard: If you go with 
```
def Basis.rightBaseChange (b : Basis ι R A) [DecidableEq ι] : Basis ι B (A ⊗[R] B) where
```
then this should be
```
lemma Basis.rightBaseChange_repr [DecidableEq ι] (b : Basis ι R A) (i) (x : B) :
```
with no `classical` or `Classical` in sight.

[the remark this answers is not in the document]
kbuzzard: Hmm, I just looked at https://leanprover-community.github.io/mathlib4_docs/Mathlib/LinearAlgebra/TensorProduct/Basis.html#Basis.baseChange and they manage to do things without assuming decidable equality at all. Maybe it's better to just copy the proof there?

FLT/Mathlib/Topology/Algebra/Module/ModuleTopology.lean
```diff
@@ -243,11 +243,13 @@ over a complete thing so I don't think there can be any other possibility
 (the argument is weak here)
 -/
 
+@[simps!]
 def continuousLinearEquiv {A B R : Type*} [TopologicalSpace A]
     [TopologicalSpace B] [TopologicalSpace R] [Semiring R] [AddCommMonoid A] [AddCommMonoid B]
     [Module R A] [Module R B] [IsModuleTopology R A] [IsModuleTopology R B]
     (e : A ≃ₗ[R] B) :
     A ≃L[R] B where
+  toFun := e
```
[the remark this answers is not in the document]
kbuzzard: Maybe they even make it more readable? I think it's fine...

FLT/Mathlib/Topology/Algebra/Valued/ValuationTopology.lean
```diff
@@ -8,3 +8,16 @@ lemma Valued.isUnit_valuationSubring_iff {Γ₀ : Type*} [LinearOrderedCommGroup
     IsUnit x ↔ Valued.v (x.val : F) = 1 := by
   convert Valuation.Integers.isUnit_iff_valuation_eq_one _
   exact Valuation.integer.integers _
+
+/-- The unit ball of a valued ring is closed. -/
+theorem Valued.integer_isClosed ( R : Type* ) {Γ₀ : Type*} [Ring R]
+    [LinearOrderedCommGroupWithZero Γ₀] [_i : Valued R Γ₀] : IsClosed (_i.v.integer : Set R) := by
```
kbuzzard, line 14: I think `_i` conventionally means "I will not ever use this variable", so maybe you want to call it `i` or `inst`?

kbuzzard (approved):

kbuzzard:
> Not the proof in the blueprint to avoid having to work with discriminants and to avoid bases as much as possible. 

That's fine!

> This also doesn't need the finite set of exceptions. 

I had not realised that this was true! I realise now what I was confused by. Here's something which is *not* true. Say M,L are K-algebras and all are number fields and if M/K is normal and "contains the Galois closure of L/K" (i.e. contains a copy of L). Then there's a natural isomorphism L tensor_K M -> sum_phi M as phi runs over the embeddings L -> M. What is not true is that this map induces an isomorphism on integer rings, because things can go wrong at ramified primes. But the point is that both L and M can be ramified at a place of K and that's where things can go wrong. In this situation where M is not a number field but a completion, it is (I guess formally?) unramified (if that makes sense in this context).

FLT/Mathlib/RingTheory/TensorProduct/Basis.lean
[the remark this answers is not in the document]
matthewjasper: They use classical here: https://github.com/leanprover-community/mathlib4/blob/446b459fb17e024ec88cdfd27021cd8b177e5e58/Mathlib/LinearAlgebra/DirectSum/Finsupp.lean#L251 .

The other problem with copying their proof is that they have  a lot of lemmas using the `leftHasSMul` instance, which we can't be used here.

FLT/DedekindDomain/AdicValuation.lean
```diff
@@ -0,0 +1,288 @@
...
+  -- Extract y by writing b as a general term of the sum of the two ideals.
+  obtain ⟨x, hx, z, hz, hxz⟩ := Submodule.mem_sup.mp hb
+  obtain ⟨y, hy⟩ := Ideal.mem_span_singleton'.mp hz
+  use y
+  -- And again prove the result about valuations by turning into one about ideals.
+  rwa [hy, ← hxz, sub_add_cancel_right, HeightOneSpectrum.intValuation_le_pow_iff_dvd,
+      Ideal.dvd_span_singleton, neg_mem_iff]
+
+open scoped Multiplicative in
+private lemma exists_ofAdd_lt_and_ofAdd_lt { x y : ℤₘ₀ } ( hx : x ≠ 0 ) ( hy : y ≠ 0) :
```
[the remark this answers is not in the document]
matthewjasper: The lemmas for this should be much better now.

FLT/DedekindDomain/AdicValuation.lean
matthewjasper: There's a lot of mathlib projects I'm think about after this.

kbuzzard:
Not that this matters, but your `force push` made `gh pr checkout 400` stop working at my end. I fixed it by deleting my version of the branch (which was not problematic).

FLT/Mathlib/Algebra/Order/GroupWithZero.lean
```diff
@@ -0,0 +1,32 @@
...
+  {x y : WithZero G} (hx : x ≠ 0) (hy : y ≠ 0) :
+    ∃ z, z ≠ 0 ∧ z ≤ x ∧ z ≤ y := by
+  let z := x ⊓ y
+  use z
+  constructor
+  . unfold z
+    intro hz
+    rw [min_eq_iff] at hz
+    tauto
+  . constructor <;> simp_all [z]
```
kbuzzard:
```suggestion
  use x ⊓ y
  refine ⟨fun hz ↦ ?_, by simp, by simp⟩
  rw [min_eq_iff] at hz
  tauto
```
No different, just noting some techniques: you don't need to introduce z, and for P \and Q \and R goals it's probably easier to use `refine \<?_, ?_, ?_\>` than two `constructors.

kbuzzard (approved):
Thanks a lot for this great effort! A bunch of this stuff should probably be in mathlib.

PR #407, opened 2025-04-13
chore(Padics): assume that `s` is a compact subgroup when computing `[xs : s]`

YaelDillies:
Otherwise it doesn't hold, as noted in https://github.com/ImperialCollegeLondon/FLT/issues/279#issuecomment-2799015518.


kbuzzard:
What about if s={0}? Don't you need it to have positive measure as well, e.g. have nonempty interior (e.g. be open)?

YaelDillies:
Hmm, yes, it needs to be both compact and open. I think Matej's suggestion to assume `s = ℤₚ` is better at this point. So I'll close this.

kbuzzard:
Thanks! I'm happy to work on Matej's approach.

PR #411, opened 2025-04-14
feat: add some distribHaarChar sorries

kbuzzard:
Need to add relevant sorries to blueprint


kbuzzard:
All subsumed by other PRs now.

PR #399, opened 2025-04-05
chore: adaptations for mathlib adele refactor

kbuzzard:
My proposed adele refactor https://github.com/leanprover-community/mathlib4/pull/23542 does some damage to the repo and means that we need to find a slightly different route for A_K tensor L = A_L . This PR picks up the pieces.

PR #413, opened 2025-04-14
chore: bump to `v4.19.0-rc3`

pitmonticone:


pitmonticone:
Ops, sorry. I'm removing it. It was not at all necessary. I just forgot it there. Thanks!

pitmonticone:
@kbuzzard I've removed the unnecessary `import Mathlib` so there is only one remaining file to fix.

kbuzzard:
This file is going to hugely break when https://github.com/leanprover-community/mathlib4/pull/23542 lands and it's already had major heart surgery in https://github.com/ImperialCollegeLondon/FLT/pull/399 so I'm not too bothered about errors here. I'll take a look now though.

pitmonticone:
Oh, I see. Do you prefer to `#exit` it temporarily?

kbuzzard:
I'm having a look now, it would be nice to get it working again. It now seems to me that the issue blocking this PR is independent from the issues which will be created by https://github.com/leanprover-community/mathlib4/pull/23542 . The first 500 lines of `FLT.DedekindDomain.FiniteAdeleRing.BaseChange` compile fine in my adele refactor branch.

kbuzzard:
Got it -- it was maxSynthPendingDepth (which I think is set to 3 by default in mathlib and probably to 1 in this repo?)

PR #417, opened 2025-04-15
feat: numeric bounds from the discriminant

b-mehta:
As discussed here: [#FLT > Odlyzko bounds @ 💬](https://leanprover.zulipchat.com/#narrow/channel/416277-FLT/topic/Odlyzko.20bounds/near/512248433) and privately.

### DIFF
--- a/FLT/NumberField/DiscriminantBounds.lean
+++ b/FLT/NumberField/DiscriminantBounds.lean
@@ -1,0 +1,212 @@
+/-
+Copyright (c) 2025 Bhavik Mehta. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Bhavik Mehta
+-/
+
+import Mathlib.Analysis.SpecialFunctions.Stirling
+import Mathlib.Data.Complex.ExponentialBounds
+import Mathlib.Data.Real.Pi.Bounds
+import Mathlib.Data.Real.StarOrdered
+import Mathlib.Tactic.NormNum.NatFactorial
+
+open scoped Nat
+open Real
+
+namespace Stirling
+
+open Filter
+
+-- upstream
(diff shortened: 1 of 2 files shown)


FLT/NumberField/DiscriminantBounds.lean
```diff
@@ -0,0 +1,174 @@
...
+      _ = (n * (n + 1)) * log n - (n + 1) * log n ! := by field_simp; ring
+      _ < n * (n + 1) * log (n + 1) - n * (log (n + 1) + log n !) := by
+          linear_combination discrBound_strictMono_aux2 n hn.ne'
+      _ = n * (n + 1) * (log (n + 1) - (n + 1 : ℝ)⁻¹ * (log (n + 1) + log n !)) := by
+          field_simp; ring
+      _ = n * (n + 1) * (log (n + 1 : ℕ) - (↑(n + 1) : ℝ)⁻¹ * log (n + 1)!) := by
+          rw [Nat.factorial_succ, Nat.cast_mul, log_mul (by positivity) (by positivity),
+            Nat.cast_add_one]
+
+noncomputable def discrBound (n : ℕ) : ℝ :=
```
kbuzzard, line 92:
```suggestion
/--
Minkowski's lower bound for the root discriminant (that is, the n'th root of the absolute value of the discriminant) of a totally complex number field of degree n.
-/
noncomputable def rootDiscrBound (n : ℕ) : ℝ :=
```

kbuzzard: I think it's better being called rootDiscrBound but other than that this looks good to go. Thanks!

kbuzzard (approved):
I think it's better being called rootDiscrBound but other than that this looks good to go. Thanks!

FLT/NumberField/DiscriminantBounds.lean
b-mehta: You'd have hated the name I gave it in the first commit then :D

kbuzzard:
Thanks so much!

PR #436, opened 2025-04-26
chore: shake project

kbuzzard:
The goal here is to make `lake exe shake FLT` work.


kbuzzard:
This is now ready for merging

FLT/Mathlib/Algebra/Module/Submodule/Basic.lean
```diff
@@ -1,3 +1,4 @@
 import Mathlib.Algebra.Module.Submodule.Basic
+-- when this file is removed, we can remove it from noshake.json
```
kbuzzard, line 2:
```suggestion
-- Once this file is upstreamed, we can remove it from noshake.json
```

PR #446, opened 2025-04-29
feat: concrete Hecke operators

kbuzzard:


kbuzzard:
This will partly be subsumed by #485 .

PR #455, opened 2025-04-30
ResidueFieldToCompletionResidueField

### LINKED ISSUE #448: Map from residue field of Dedekind domain to residue field of completion
If A is a Dedekind domain with field of fractions K, and if v is in the height one spectrum of A (i.e. is the valuation associated to a nonzero prime ideal P_v of A), and if O_v denotes the integer ring of the completion K_v, then we have that O_v is local and we have the canonical map A -> O_v. The claim is that this induces a map A / P_v -> residue field of O_v. This boils down to knowing that the maximal ideal of O_v is precisely the stuff with valuation < 1 (which we may or may not have), that v on K_v agrees with v on A (which I'm pretty sure we do have) and that P_v is the stuff in A with valuation less than 1 (which surely we have).

Note: the issue here is just defining the map. The next issue is the claim that it's an isomorphism.

pechersky:
Closes #448


kbuzzard:
Thanks!

PR #474, opened 2025-05-09
Dimension results for extensions of infinite place completions

smmercuri:
If `w` is an infinite place of `L/K` lying above `v : InfinitePlace K` then this PR proves:
- the dimension of `w.Completion` over `v.Completion` is equal to the ramification index -- `finrank_eq_ramificationIdx`;
- the sum of all ramification indices over `v` is `[L : K]` -- `sum_ramificationIdx_eq`;
- the `v.Completion`-dimension of `(w : v.Extension L) → w.1.Completion` equals that of `L ⊗[K] v.Completion` (both `[L : K]`) -- `finrank_pi_eq_finrank_tensorProduct`


FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+- `v` is real, `w` is complex (`w` is a ramified extension of `v`), and
+  both `w.embedding` and `conjugate w.embedding` extend `v.embedding`.
+- `v` and `w` are both real (i.e., `w` is an unramified real extension of `v`) and
+  `w.embedding` extends `v.embedding`.
+- `v` and `w` are both complex (`w` is an unramified complex extension of `v`), and
+  `w.embedding` extends `v.embedding`.
+- `v` and `w` are both complex, and `conjugate w.embedding` extends `v.embedding`.
+
+## Main definitions
+- `ComplexEmbedding.IsExtension f g` : predicate determining whether `g : L →+* ℂ`
```
kbuzzard, line 22:
```suggestion
- `ComplexEmbedding.IsExtension f g` : predicate asserting that `g : L →+* ℂ`
```

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+  extends `f : L →+* ℂ`.
+- `ComplexEmbedding.IsMixedExtension f g` : predicate determining whether `g : L →+* ℂ` extends
```
kbuzzard, line 24:
```suggestion
- `ComplexEmbedding.IsMixedExtension f g` : predicate asserting that `g : L →+* ℂ` extends
```

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+  `f : L →+* ℂ` and `g` is complex while `f` is real.
+- `ComplexEmbedding.IsUnmixedExtension f g` : predicate determining whether `g : L →+* ℂ` extends
+  `f : L →+* ℂ` and `g` is real if and only if `f` is real.
```
kbuzzard, line 27:
```suggestion
- `ComplexEmbedding.IsUnmixedExtension f g` : predicate asserting `g : L →+* ℂ` extends
  `f : L →+* ℂ` and that `g` is real if and only if `f` is real.
```

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+  let ⟨x, hx⟩ := hr
+  exact ⟨algebraMap K L x, by aesop⟩
+
+/--
+If `L/K` and `f : K →+* ℂ`, `g : L →+* ℂ`, then `g` is a _mixed extension_ of `f` if the
+image of `f` is real while the image of `g` is complex.
+
+This is the complex embedding analogue of `InfinitePlace.RamifiedExtension`.
+-/
+abbrev IsMixedExtension :=
```
kbuzzard, line 81: So people don't just call this a ramified extension?

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+end IsMixedExtension
+
+/--
+If `L/K` and `f : K →+* ℂ`, `g : L →+* ℂ`, then `g` is an _unmixed extension_ of `f` if `g` is an
+extension of `f` but is not a mixed extension. In other words, the image of `f` is real
+if and only if the image of `g` is real.
+
+This is the complex embedding analogue of `InfinitePlace.UnramifiedExtension`.
+-/
+abbrev IsUnmixedExtension := IsExtension f g ∧ ¬IsMixedExtension f g
```
kbuzzard, line 109: So people don't just call this an unramified extension?

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+    w.comap f x = w (f x) := rfl
+
+variable {v w} in
+theorem comp_of_comap_eq {f : K →+* L} (h : w.comap f = v) (x : K) :
+    w (f x) = v x := by
+  simp [← h]
+
+/-- Class encoding the fact that an infinite place is real. -/
+class Real where
+  isReal' : v.IsReal
```
kbuzzard, line 152: Should we delete `NumberField.InfinitePlace.IsReal` then?

FLT/NumberField/InfinitePlace/Extension.lean
```diff
@@ -0,0 +1,334 @@
...
+
+theorem Real.isReal [v.Real] : v.IsReal := Real.isReal'
+
+/-- Class encoding the fact that an infinite place is complex. -/
+class Complex where
+  isComplex' : v.IsComplex
+
+theorem Complex.isComplex [v.Complex] : v.IsComplex := Complex.isComplex'
+
+theorem not_real_iff_complex : ¬v.Real ↔ v.Complex := by
```
kbuzzard, line 162: If you're going to have lemmas like this then maybe you should prove v.Real <-> v.IsReal and the same for Complex, and then just rewrite those and `not_isReal_iff_isComplex` -- no need to split into two cases. But if you want lemmas like this then why make `NumberField.InfinitePlace.Real` at all, that's not usually how you use classes.

FLT/NumberField/InfinitePlace/Extension.lean
smmercuri: Hmm I was a bit unsure, and wanted to avoid ramified/unramified since they are already used for infinite places, but maybe it's ok to re-use the terms? I guess we are already using the word "extension" for both

FLT/NumberField/InfinitePlace/Extension.lean
smmercuri: I think my original thinking was that we would turn `IsReal` into a class when upstreaming, but maybe that's not worth the hassle to refactor. Probably makes more sense to use `[Fact v.IsReal]` where needed, so I've updated the PR to do that

FLT/NumberField/InfinitePlace/Extension.lean
kbuzzard: Aah I see, I had missed the distinction between places and embeddings. Yeah let's just leave it as it is.

FLT/NumberField/InfinitePlace/Extension.lean
kbuzzard: I now understand the distinction you're drawing.

FLT/NumberField/InfinitePlace/Extension.lean
kbuzzard: Have you?

smmercuri:
Oh yeah I think it's because that PR had modified a proof in `Completion.Infinite`. Also now updated the blueprint in this PR to reflect the name change `InfinitePlace.ExtensionPlace` -> `InfinitePlace.Extension`.

PR #475, opened 2025-05-09
Fill in definition of NumberField.AdeleRing.baseChangeEquiv

matthewjasper:
I'm leaving continuity (well, openness) for a future PR

### DIFF
--- a/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
+++ b/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
@@ -987,6 +987,9 @@ noncomputable
 instance BaseChange.algebra : Algebra (FiniteAdeleRing A K) (FiniteAdeleRing B L) :=
   RingHom.toAlgebra (FiniteAdeleRing.mapRingHom A K L B)
 
+lemma FiniteAdeleRing.mapSemialgHom_continuous : Continuous (mapSemialgHom A K L B) :=
+  sorry
+
 attribute [instance 100] RestrictedProduct.instSMulCoeOfSMulMemClass
 -- otherwise
 -- #synth SMul (FiniteAdeleRing A K) (FiniteAdeleRing B L)
@@ -1014,5 +1017,16 @@ noncomputable def FiniteAdeleRing.baseChangeContinuousAlgEquiv :
   continuous_toFun := sorry
   continuous_invFun := sorry
   -- TODO needs issue number
+
+noncomputable instance baseChangeAlgebra : Algebra K (FiniteAdeleRing B L) :=
+  RingHom.toAlgebra <| (algebraMap L _).comp (algebraMap K L)
+
+noncomputable instance baseChangeScalarTower :
+    IsScalarTower K (FiniteAdeleRing A K) (FiniteAdeleRing B L) := by
+  apply IsScalarTower.of_algebraMap_eq
+  intro x
+  nth_rw 2 [RingHom.algebraMap_toAlgebra]
+  symm
+  exact SemialgHom.commutes (FiniteAdeleRing.mapSemialgHom A K L B) x
 
 end IsDedekindDomain
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -135,6 +135,15 @@ local instance : IsModuleTopology (FiniteAdeleRing (𝓞 K) K) (D ⊗[K] (Finite
 
 variable [FiniteDimensional K D]
 
+-- Instance to help speed up instance synthesis
+instance : NonUnitalNonAssocRing (D ⊗[K] (FiniteAdeleRing (𝓞 K) K)) :=
+  let r := Algebra.TensorProduct.instRing.toNonUnitalRing
+  r.toNonUnitalNonAssocRing
+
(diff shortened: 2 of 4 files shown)


FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -135,6 +135,15 @@ local instance : IsModuleTopology (FiniteAdeleRing (𝓞 K) K) (D ⊗[K] (Finite
 
 variable [FiniteDimensional K D]
 
+-- Instance to help speed up instance synthesis
```
matthewjasper, line 138: I wasn't able to find a better solution to instances timing out here

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -1015,4 +1018,15 @@ noncomputable def FiniteAdeleRing.baseChangeContinuousAlgEquiv :
   continuous_invFun := sorry
   -- TODO needs issue number
 
+noncomputable instance baseChangeAlgebra : Algebra K (FiniteAdeleRing B L) :=
+  RingHom.toAlgebra <| (algebraMap L _).comp (algebraMap K L)
```
kbuzzard, line 1022: Do you think we should be making all of these Algebra instances? Initially I was doing this but then I switched to semialgebra homs. My motivation was that I wanted to say things like "the natural map from A_K to A_L is more than just a ring homomorphism, it preserves some other structure" and initially I was saying "it's a K-algebra homomorphism" but now I can say "it's a semialgebra morphism over the natural map K -> L" and this avoids having to make A_L a K-algebra.

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: I think it's fine to make instance shortcuts. We're pushing typeclass inference to the limits it seems with these arguments. As long as there are no diamonds I think shortcuts are fine.

FLT/NumberField/AdeleRing.lean
```diff
@@ -30,26 +34,47 @@ section BaseChange
...
 
 instance : IsScalarTower K L (𝔸 L) :=
-  IsScalarTower.of_algebraMap_eq' rfl
+  IsScalarTower.of_algebraMap_eq fun _ ↦ rfl
 
 /-- The canonical map from the adeles of K to the adeles of L -/
 noncomputable def baseChange :
     (𝔸 K) →A[K] 𝔸 L :=
-  sorry -- product of finite and infinite adele maps
+  let finite : FiniteAdeleRing (𝓞 K) K →A[K] FiniteAdeleRing (𝓞 L) L := {
```
kbuzzard, line 52: Do you think that really this should be a continuous semialgebra hom over K -> L?

FLT/NumberField/AdeleRing.lean
```diff
@@ -58,12 +83,41 @@ instance : IsModuleTopology (𝔸 K) (L ⊗[K] 𝔸 K) := ⟨rfl⟩
 instance instPiIsModuleTopology : IsModuleTopology (𝔸 K) (Fin (Module.finrank K L) → 𝔸 K) :=
   IsModuleTopology.instPi
 
-open IsDedekindDomain in
+instance instBaseChangeIsModuleTopology : IsModuleTopology (𝔸 K) (𝔸 L) := by
+  sorry
```
kbuzzard, line 87: My plan here is to show that this follows from the corresponding statement at the finite and infinite places, and that the claim at the finite places follows from the fact that A_L^infty is isomorphic (both topologically and algebraically) to the restricted product over the finite places v of *K* of (prod_{w|v} L_w) with respect to (prod_{w|v} O_w) and then to say that prod_{w|v}L_w has the K_v-module topology, prod_{w|v}O_w has the O_v-module topology, and then hope that there's a general theorem about module topologies of restricted products.

FLT/NumberField/InfiniteAdeleRing.lean
```diff
@@ -137,4 +137,11 @@ def baseChangeEquiv :
   IsModuleTopology.continuousAlgEquivOfIsScalarTower K _ (baseChangeEquivAux K L)
     (by simp_rw [baseChangeEquivAux_apply]; exact SemialgHom.baseChange_of_algebraMap_tmul_right _)
 
+instance : IsScalarTower K (InfiniteAdeleRing K) (InfiniteAdeleRing L) :=
```
kbuzzard, line 140: Again, do you think that we should even have InfiniteAdeleRing L being a K-algebra?

kbuzzard:
I'm not saying "don't make the algebra instances", I'm really just looking for your opinion.

FLT/NumberField/AdeleRing.lean
matthewjasper: For the purposes of defining the base change isomorphism for ideals, this would be fine as a normal semialgebra homomorphism (and a lemma that it's continuous if we want that).

FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
kbuzzard: Oh that's funny! So we need that L tensor_K A_K -> A_L is L-linear, K-linear and A_K-linear! No that's fine, you've convinced me that this is useful.

PR #482, opened 2025-05-11
Kbuzzard fix conflicting defs

kbuzzard:
In `FLT/Patching/Utils/AdicTopology.lean` we have a definition `isOpen_maximalIdeal_pow` which doesn't demand a compactness hypothesis. Unfortunately in `Mathlib.Topology.Algebra.Ring.Compact` from mathlib there is also a definition of `isOpen_maximalIdeal_pow` and this one does. This means we can't import Mathlib and FLT at the same time. 

Currently I've fixed this by adding `import Mathlib.Topology.Algebra.Ring.Compact` to `FLT/Patching/Utils/AdicTopology.lean` and then renaming the conflicting names with primes (I needed two primes for `isOpen_maximalIdeal_pow''` as there's also a primed version in the same FLT file).

Similar problems occur with `isOpen_maximalIdeal` and `isOpen_iff_finite_quotient` (the mathlib versions seem to be different to our versions), but these do not seem to be used.


erdOne (approved):
Yeah most things in `FLT/Patching/Utils/AdicTopology.lean` are actually not used but they are interesting (maybe not to FLT but to mathlib) in their own rights.

The changes look right to me.

PR #489, opened 2025-05-12
Split base change

matthewjasper:
Splits the file so that the non-adele defs/theorems are in separate files


kbuzzard:
In mathlib there's a tool which confirms nothing got lost in such PRs but we don't seem to have it here.

PR #497, opened 2025-05-14
Bump mathlib

Ruben-VandeVelde:
This also fixes the warning "manifest out of date: git revision of dependency 'mathlib' changed; use `lake update mathlib` to update it" by correcting the inputRev field.


kbuzzard:
(sorry, I accidentally pushed an `import Mathlib` and now CI takes over an hour because it wastes its time building a gazillion mathlib docs; fixed in #498 )

PR #498, opened 2025-05-14
chore: shake

kbuzzard:
I accidentally pushed an `import Mathlib` and suddenly docs took 45 minutes to build aargh.

@YaelDillies you have a solution for this right?

Would another solution be "get CI to shake"? This might be preferred (mathlib does this, right?) Is this easy?


YaelDillies:
Yes, the following CI step fails if some `import Mathlib` is present:
https://github.com/YaelDillies/LeanAPAP/blob/master/.github/workflows/push_master.yml#L14-L27

PR #505, opened 2025-05-17
latex-workshop tweaks

PatrickMassot:


kbuzzard:
Well, it works on my laptop! Here's hoping that this is the last of the problems I face with LaTeX workshop.

@pitmonticone you might want to consider putting this in your template.

Thanks a lot Patrick.

pitmonticone:
Sure! I'll do.

Does `leanblueprint pdf` still work properly?

pitmonticone:
PS: Only the `.vscode/settings.json` file can be added to the `LeanProject` template. The other one (`blueprint/src/latexmkrc`) should be change in the `LeanBlueprint` template.

PR #534, opened 2025-05-21
feat: mulEquivHaarChar_trans

Ruben-VandeVelde:


kbuzzard:
@Ruben-VandeVelde do you mind if I merge #535 instead? Remy's proof is better than the proof in the blueprint.

PR #536, opened 2025-05-21
chore: try to stop CI randomly taking an hour

kbuzzard:
#523 took nearly an hour in CI and I don't know why. More precisely https://github.com/ImperialCollegeLondon/FLT/actions/runs/15141983212/job/42568200037 shows that it spent 42 minutes building API docs, even though we didn't bump mathlib. Bhavik suggests that this PR might fix this.

@pitmonticone what do you think? I guess we're still figuring out how to run doc-gen in the preferred way.

### DIFF
--- a/scripts/build_docs.sh
+++ b/scripts/build_docs.sh
@@ -39,7 +39,11 @@ sed -i s/TOOLCHAIN/`grep -oP 'v4\..*' lean-toolchain`/ docbuild/lakefile.toml
 cd docbuild
 
 # Disable an error message due to a non-blocking bug. See Zulip
-MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update FLT
+# (note added by kmb : no idea what that "See Zulip" is referring to but Bhavik
+# suggests that we should be updating doc-gen4 rather than FLT here in order to prevent
+# the "build project API documentation" step of CI from taking 42 minutes on
+# e.g. PR 523 despite not bumping mathlib)
+MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update doc-gen4
 
 # Build the docs
 ~/.elan/bin/lake build FLT:docs


kbuzzard:
The "build project API documentation" only took 7 minutes on this one. But then again it only took 8 minutes on #527 . However CI took 54 minutes on #525 with building docs being 42 minutes of it, despite the fact that the PR does not change any imports at all. 

Is the philosophy with CI PRs just "merge it and see what happens, maybe it will stop CI randomly taking 54 minutes in which case yay, and maybe it won't in which case try something else"?

pitmonticone:
I don't think `lake update <library>` is ever necessary in our use cases. `MATHLIB_NO_CACHE_ON_UPDATE=1 ~/.elan/bin/lake update` should suffice.

scripts/build_docs.sh
```diff
@@ -39,7 +39,11 @@ sed -i s/TOOLCHAIN/`grep -oP 'v4\..*' lean-toolchain`/ docbuild/lakefile.toml
 cd docbuild
 
 # Disable an error message due to a non-blocking bug. See Zulip
```
b-mehta, line 41: imo this message isn't at all helpful, and we should replace it with a link to something like this: https://github.com/leanprover/doc-gen4?tab=readme-ov-file#usage since point 3 is the one saying to do this

pitmonticone: I agree. I didn't add that btw :)

pitmonticone:
> Testing it here [fpvandoorn/carleson#346](https://github.com/fpvandoorn/carleson/pull/346)

It worked fine with the usual 7-8 minutes.

scripts/build_docs.sh
b-mehta: I had no intention of blaming you :)

pitmonticone:
To summarise: I think the configuration we have in Carleson now is basically what we want in FLT too https://github.com/fpvandoorn/carleson/blob/master/scripts/build_docs.sh

b-mehta:
Do you have an idea why CI takes a full hour (on PRs that don't bump mathlib) occasionally? It's not consistent, so it's hard to be sure if one particular PR affects it

pitmonticone:
No, it requires further investigation. I would probably find a bit of time during the weekend.

kbuzzard:
But #525 adds no imports, it adds two new lemmas and fills in some sorries.

pitmonticone:
> But #525 adds no imports, it adds two new lemmas and fills in some sorries.

And it takes 7-8 minutes, right?

pitmonticone:
Yes, but that could just be because in their fork they had the `import Mathlib` you removed about 40 minutes after their 4th commit.

pitmonticone:
So, at the end of the day, I wouldn’t be concerned about this. The simplification of `build_docs.sh`, including golfing and using `lake update` instead of `lake update FLT` is a nice improvement, but I don’t believe it fixes anything, as there doesn’t seem to be anything that actually needed fixing.

pitmonticone:
Finally, in [this comment](https://github.com/ImperialCollegeLondon/FLT/pull/525#issuecomment-2898091555) I asked William to avoid developing in the main branch to avoid similar issues in the future.

b-mehta:
> Finally, in [this comment](https://github.com/ImperialCollegeLondon/FLT/pull/525#issuecomment-2898091555) I asked William to avoid developing in the `main` branch to avoid similar issues in the future.

What is it about `main` branches that causes this? Can't we adjust the pipeline (in particular the if blocks) to allow this somehow?

pitmonticone:
The point about the `main` branch in forks is simply to avoid confusion for investigation and debugging purposes. 

It would have been much clearer that he was using the version with `import Mathlib` without small those pushing and pulling.

PR #538, opened 2025-05-21
feat: prove ringHaarChar_mul_volume

### LINKED ISSUE #515: Measure of uX is fudge factor times measure of X
This is `ringHaarChar_mul_volume`. It should follow immediately from `addEquivAddHaarChar_smul_preimage` applied to u^{-1}X.

RemyDegenne:
Also golf ringHaarChar_mul_integral.

Closes #515


kbuzzard:
Thanks a lot for contributing to the formalized proof of Fermat's Last Theorem!

PR #540, opened 2025-05-21
disable generating equations

b-mehta:
This is an experiment to test if and how fast CI works if equations are disabled in doc-gen. 
Not ready for merging.


kbuzzard:
Wooah it took ages. Now push a blank line commit and see if it's quick?

b-mehta:
It's possible that caching doesn't work properly because it's using a modified doc-gen, and that's getting rebuilt from scratch each time

kbuzzard:
I'm happy to not build the docs at all, so I'm closing in favour of #573 and #541

PR #543, opened 2025-05-22
feat: isHaarMeasure_comap_of_isOpenEmbedding

### LINKED ISSUE #507: Pullback of Haar measure along an isomorphism
The pullback of Haar measure along a homeomorphism is a Haar measure. This should be relatively straightforward.

DavidLedvinka:
null

Closes #507


DavidLedvinka:
awaiting-review

kbuzzard:
If both this PR and #544 are using the same lemmas then do you want to make just one PR closing both issues? Don't worry about the whole task claiming thing, I can manage all this manually, but would it be easier for you to just close this one and add its results to #544? I've left more extensive comments there about how to manage lemmas which should really be in mathlib.

PR #548, opened 2025-05-22
Fujisaki : WIP Fujisaki

WilliamCoram:
First attempt at compact_quotient proof (Theorem 10.1). Think I have a method modulo sorries, feedback appreciated.

### DIFF
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -254,8 +254,25 @@ abbrev Dfx := (D ⊗[K] (FiniteAdeleRing (𝓞 K) K))ˣ
 noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
   Units.map Algebra.TensorProduct.includeLeftRingHom.toMonoidHom
 
+-- its breaking?? not sure what the difference to above is
+-- NumbeField.AdeleRing vs FiniteAdeleRing
+def α : (ringHaarChar_ker D_𝔸 ⧸ (MonoidHom.range
+    (NumberField.AdeleRing.DivisionAlgebra.incl K D)).comap (ringHaarChar_ker D_𝔸).subtype)
+    → (Dfx K D ⧸ (incl₁ K D).range) :=
+  fun a => a -- not sure if this is correct
+
 theorem NumberField.FiniteAdeleRing.DivisionAlgebra.units_cocompact :
     CompactSpace (Dfx K D ⧸ (incl₁ K D).range) := by
+  have h1 : Continuous α := by
+    -- need to give quotient topology on RHS, then 'readily verified'
+    sorry
+  have h2 : IsSurjective α := by
+    -- main part of the proof, may need to be seperate lemma
+    sorry
+  have h3 : Set.preimage α Set.univ = Set.univ := by
+    -- surjective
+    sorry
+  -- exact IsCompact.image α (compact_quotient K D) h1 -- should work?
   sorry
 
 -- Voight "Main theorem 27.6.14(b) (Fujisaki's lemma)"
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -169,8 +169,49 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
 
 end Aux
 
+def incl₂ : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  fun i => (i.1, i⁻¹.1)
+
+def M : Set (ringHaarChar_ker D_𝔸) := Set.preimage (incl₂ K D) (Aux.C K D)
+
+def MtoQuot : (ringHaarChar_ker D_𝔸) → (ringHaarChar_ker D_𝔸 ⧸
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) :=
+  fun a => Quot.mk _ a
+
+lemma rinHaarChar_ker_isCompact : IsCompact (ringHaarChar_ker D_𝔸) := by
+  -- this is true since this is a closed subset of D_𝔸ˣ, which is a closed subset of D_𝔸 x D_𝔸
+  -- which is compact by product of compact spaces
+  sorry
+
 lemma compact_quotient : CompactSpace (ringHaarChar_ker D_𝔸 ⧸
-  (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := sorry
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := by
+  have h1 : IsClosed (M K D) := by
+    have h11 : Continuous (incl₂ K D) := by
+      -- inclusion? so should be trivial by product topology (and subspace topology)
+      sorry
+    have h12 : IsClosed (Aux.C K D) := by
+      -- I think this is what I need?
+      sorry
+    rw [M]
+    exact IsClosed.preimage h11 h12
+  have h2 : IsCompact (M K D) := by
+    -- exact IsClosed.isCompact h1
+    sorry
+  refine isCompact_univ_iff.mp ?_
+  have h3 : Set.SurjOn (MtoQuot K D) (M K D) Set.univ := by
+    sorry
+  have h4 : Continuous (MtoQuot K D) := by
+    exact { isOpen_preimage := fun s a ↦ a }
+  have h5 : IsClosed (Set.preimage (MtoQuot K D) (Set.univ)) := by
+    exact closure_subset_iff_isClosed.mp fun ⦃a⦄ a ↦ trivial
+  have h6 : IsCompact (Set.preimage (MtoQuot K D) (Set.univ)) := by
+    -- exact IsClosed.isCompact h5
+    sorry
+  have h7 := IsCompact.image h6 h4
+  have h8 : (MtoQuot K D '' (MtoQuot K D ⁻¹' Set.univ)) = Set.univ := by
+    exact Set.SurjOn.image_preimage h3 fun ⦃a⦄ a ↦ a
+  rw [h8] at h7
+  exact h7
 
 end NumberField.AdeleRing.DivisionAlgebra
 
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -205,6 +205,7 @@ lemma compact_quotient : CompactSpace (ringHaarChar_ker D_𝔸 ⧸
   have h5 : IsClosed (Set.preimage (MtoQuot K D) (Set.univ)) := by
     exact closure_subset_iff_isClosed.mp fun ⦃a⦄ a ↦ trivial
   have h6 : IsCompact (Set.preimage (MtoQuot K D) (Set.univ)) := by
+    rw [IsCompact]
     -- exact IsClosed.isCompact h5
     sorry
   have h7 := IsCompact.image h6 h4
@@ -266,6 +267,26 @@ of `(D ⊗ 𝔸_F^infty)ˣ`.
 theorem NumberField.FiniteAdeleRing.DivisionAlgebra.finiteDoubleCoset
     {U : Subgroup (Dfx K D)} (hU : IsOpen (U : Set (Dfx K D))) :
     Finite (Doset.Quotient (Set.range (incl₁ K D)) U) := by
+  have ToFinCover := isCompact_univ_iff.mpr
+    (NumberField.FiniteAdeleRing.DivisionAlgebra.units_cocompact K D)
+  apply isCompact_iff_finite_subcover.mp at ToFinCover
+  have openCover :  (Doset.Quotient (Set.range (incl₁ K D)) U) = ⋃ (q : Dfx K D),
+      Doset.doset q (Set.range (incl₁ K D)) U := by
+    -- this should be true by definition (if I am reading them right)
+    -- if not I still want something like this; as this is the open cover we will consider
+    sorry
+  rw [openCover]
+  refine Set.finite_coe_iff.mpr ?_
+  have := Doset.doset_union_rightCoset (incl₁ K D).range U
+  simp only [MulOpposite.op_mul, MonoidHom.coe_range] at this
+  simp_rw [← this]
+  -- The hope is I can show the inner union is finite (by reducing open cover)
+  -- need to work out if it is rightCoset or leftCoset I want though...
+  -- there may be a better way to be doing this though
+
+  -- e.g constructing map from doset to single (left or rights, IDK the naming convention) cosets
+  -- then show the set of these are finite (reducing cover)
+  -- then push this back up and show it covers
   sorry
 
 end FiniteAdeleRing
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -132,8 +132,6 @@ lemma X_meets_kernel' {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
 /-- An auxiliary set T used in the proof of Fukisaki's lemma. Defined as Y ∩ Dˣ. -/
 def T : Set D_𝔸ˣ := ((↑) : D_𝔸ˣ → D_𝔸) ⁻¹' (Y K D) ∩ Set.range ((incl K D : Dˣ → D_𝔸ˣ))
 
--- Need something saying D ⊆ D_𝔸 is discrete
-
 lemma T_finite : Set.Finite (T K D) := by
   have h : Set.Finite ((Y K D) ∩ (Set.range (Algebra.TensorProduct.includeLeft : D →ₐ[K] D_𝔸)))
       := by
@@ -142,8 +140,8 @@ lemma T_finite : Set.Finite (T K D) := by
 
       -- Subgroup.isClosed_of_discrete
       sorry
-    · -- follows form D being discrete
-
+    ·
+      -- follows form D being discrete
       sorry
   have h1 : Units.val '' T K D ⊆ (Y K D) ∩
       (Set.range (Algebra.TensorProduct.includeLeft : D →ₐ[K] D_𝔸)) := by
@@ -186,12 +184,22 @@ lemma compact_quotient : CompactSpace (ringHaarChar_ker D_𝔸 ⧸
   have h2 : IsCompact (M K D) := by
     apply Topology.IsClosedEmbedding.isCompact_preimage
     · refine Topology.IsClosedEmbedding.of_continuous_injective_isClosedMap ?_ ?_ ?_
-      · -- need to show continuity
+      · rw [incl₂]
+        refine continuous_induced_rng.mp ?_
+        have := MeasureTheory.ringHaarChar_continuous D_𝔸
+        -- think it is an application of this somewhere?
         sorry
-      · -- need to show injectivity
-        sorry
-      · -- not sure if this is the same as closed immersion?
-        sorry
+      · intro a b eq
+        simp_rw [incl₂] at eq
+        aesop
+      · refine Topology.IsInducing.isClosedMap ?_ ?_
+        · -- true by definition?
+          sorry
+        · -- true as ringHaarChar_ker is closed... so their inclusions i, i⁻¹ are closed
+          simp_rw [incl₂]
+          simp only [Subgroup.coe_subtype]
+
+          sorry
     · exact Aux.C_compact K D
   have h3 : (MtoQuot K D) '' (M K D) = Set.univ := by
     ext x
@@ -247,11 +255,13 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
   Units.map Algebra.TensorProduct.includeLeftRingHom.toMonoidHom
 
 variable [MeasurableSpace (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)]
-  [BorelSpace (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)] -- not sure why I need to restate these from
-  -- the start of the file?
+  [BorelSpace (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)]
 
 def rest₁ : ringHaarChar_ker D_𝔸 → Dfx K D :=
-  -- this should just be the restriction after removing infinite places
+  fun a => --((Subgroup.subtype (ringHaarChar_ker D_𝔸)) a)
+
+           -- the RHS is tensoring over all places; just need to work out how to remove the infinite
+           -- ones... this should be a .2 somewhere; not sure how to do this restriction in Lean
   sorry
 
 def α : (ringHaarChar_ker D_𝔸 ⧸ (MonoidHom.range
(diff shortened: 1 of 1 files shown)


WilliamCoram:
Now included some ideas for the finitedoublecoset proof

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,50 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
 end Aux
 
+def incl₂ : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  fun i => (i.1, i⁻¹.1)
+
+def M : Set (ringHaarChar_ker D_𝔸) := Set.preimage (incl₂ K D) (Aux.C K D)
+
+def MtoQuot : (ringHaarChar_ker D_𝔸) → (ringHaarChar_ker D_𝔸 ⧸
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) :=
+  fun a => Quot.mk _ a
```
kbuzzard, line 179: This already has a name, namely Quot.mk . You might want to avoid making too many definitions, they clog the place up. Have you considered `abbrev` instead? Quot.mk will have a ton of API and if you make MtoQuot as a def then it's suddenly much harder to use any of it because Lean won't unfold the definition. If you make it an abbrev then Lean just prints it as MtoQuot but in the actual term it's still Quot.mk so things like simp will work much better.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,50 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+      -- I think this is what I need?
+      sorry
+    rw [M]
+    exact IsClosed.preimage h11 h12
+  have h2 : IsCompact (M K D) := by
+    -- exact IsClosed.isCompact h1
+    sorry
+  refine isCompact_univ_iff.mp ?_
+  have h3 : Set.SurjOn (MtoQuot K D) (M K D) Set.univ := by
+    sorry
```
kbuzzard, line 202: I'm not sure this buys you any more than `(MtoQuot K D) '' (M K D) = Set.univ` which might be easier to work with.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,50 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+  have h4 : Continuous (MtoQuot K D) := by
+    exact { isOpen_preimage := fun s a ↦ a }
+  have h5 : IsClosed (Set.preimage (MtoQuot K D) (Set.univ)) := by
+    exact closure_subset_iff_isClosed.mp fun ⦃a⦄ a ↦ trivial
+  have h6 : IsCompact (Set.preimage (MtoQuot K D) (Set.univ)) := by
+    rw [IsCompact]
+    -- exact IsClosed.isCompact h5
+    sorry
```
kbuzzard, line 210: I doubt this is true, this just seems to be the claim that D_A^(1) is compact, which is false in general (it's probably always false in fact).

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -212,8 +254,25 @@ abbrev Dfx := (D ⊗[K] (FiniteAdeleRing (𝓞 K) K))ˣ
 noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
   Units.map Algebra.TensorProduct.includeLeftRingHom.toMonoidHom
 
+-- its breaking?? not sure what the difference to above is
+-- NumbeField.AdeleRing vs FiniteAdeleRing
```
kbuzzard, line 258: If you need a topology on `D ⊗[K] NumberField.AdeleRing (𝓞 K) K` then make it an algebra over `NumberField.AdeleRing (𝓞 K) K` and then just give it the `NumberField.AdeleRing (𝓞 K) K`-module topology.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,50 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+      -- inclusion? so should be trivial by product topology (and subspace topology)
+      sorry
+    have h12 : IsClosed (Aux.C K D) := by
+      -- I think this is what I need?
+      sorry
+    rw [M]
+    exact IsClosed.preimage h11 h12
+  have h2 : IsCompact (M K D) := by
+    -- exact IsClosed.isCompact h1
+    sorry
```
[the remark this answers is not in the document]
WilliamCoram: I think I have confused myself a bit here. Just to be clear, what compact space is M a subspace of?
Also it seems closed immersions are defined via quivers in mathlib - is using closed embeddings the same?

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: I realise this is the exposition from the start of the file - thus I have just removed the ``local``  part on the needed instances. Hopefully, this is an okay thing to do

FLT/DivisionAlgebra/Finiteness.lean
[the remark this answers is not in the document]
kbuzzard: M is a subset of C. Oh I don't mean the algebraic geometry closed immersions, I am just talking topology, I mean X -> Y is a closed immersion if it's an injection, the image is closed, and the topology on X is just the pullback along the map of the one on Y (so a subset of X is closed iff the corresponding subset of Y is closed). Yeah looks like it's [Topology.IsClosedEmbedding](https://leanprover-community.github.io/mathlib4_docs/Mathlib/Topology/Defs/Induced.html#Topology.IsClosedEmbedding)

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: I think that if you remove the `local` then the instances will permeate through every file which imports this file, and I'm not sure we want this (although I also can't imagine what other topology one could put on that object). If you need the instances elsewhere in the file just move them out of the namespace? Local instances are scoped to namespaces/sections so the ones I put there are just in the namespace which I opened on line 37 and closed on line 193.. Just move them out of the namespace and keep them local if you want them elsewhere in the file.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -12,6 +12,8 @@ import Mathlib.Algebra.Central.Defs
 import FLT.NumberField.AdeleRing
 import FLT.HaarMeasure.HaarChar.Ring
 
+set_option maxHeartbeats 1000000
```
kbuzzard, line 15: I'd be reluctant to merge this -- it's better to figure out what is slow and to fix it. In fact it doesn't seem to do anything -- you have two errors in this PR whether or not you set this option. I would advise against pushing errors -- they will ensure that CI fails. Just sorry stuff and write erroring code in comments rather than pushing errors.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -140,8 +140,8 @@ lemma T_finite : Set.Finite (T K D) := by
 
       -- Subgroup.isClosed_of_discrete
       sorry
-    · -- follows form D being discrete
-
+    ·
+      -- follows form D being discrete
```
kbuzzard, line 144: from

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+def incl₂3 : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  fun u => (unitsEquivProdSubtype D_𝔸) (Subgroup.subtype (ringHaarChar_ker D_𝔸) u)
+
+def incl₂2 : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  fun u => let p := (Units.embedProduct D_𝔸) (Subgroup.subtype (ringHaarChar_ker D_𝔸) u)
+           (p.1, MulOpposite.unop p.2)
+
+def incl₂ : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  (fun i => (i.1, i⁻¹.1)).comp (Subgroup.subtype (ringHaarChar_ker D_𝔸))
+
```
kbuzzard, line 181:
```
example : incl₂3 = incl₂2 := rfl
example : incl₂3 = incl₂ := rfl
```
They're all definitionally equal

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+abbrev MtoQuot : (ringHaarChar_ker D_𝔸) → (ringHaarChar_ker D_𝔸 ⧸
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := Quot.mk _
+
 lemma compact_quotient : CompactSpace (ringHaarChar_ker D_𝔸 ⧸
-  (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := sorry
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := by
+  have h4 : Continuous (MtoQuot K D) := by
+    exact { isOpen_preimage := fun s a ↦ a }
+  have h2 : IsCompact (M K D) := by
```
kbuzzard, line 191: Why prove this here in the middle of a much bigger proof? Lean wants lots of short proofs, not one huge one. If you need compactness of M, prove it after the definition of M. And if you need that one of your incl2's is a closed embedding prove that after you've defined incl2. Every definition you make comes with a cost, and the cost is that you have to make the API for that definition, to make it usable. But you *definitely* don't want to be making the API in the middle of some other proof of something else! That way you can't reuse your work, and it makes proofs less modular and longer.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+    apply Topology.IsClosedEmbedding.isCompact_preimage
+    · refine Topology.IsClosedEmbedding.of_continuous_injective_isClosedMap ?_ ?_ ?_
```
kbuzzard, line 193: Your goal at this point is `IsClosedEmbedding (incl₂3 K D)`. I am imagining that `ringHaarChar_ker R` (which is a term of type `Subgroup Rˣ`), when treated as a type, just inherits the subspace topology from `Rˣ`. So we have two maps here, one from the kernel to Rˣ and one from `Rˣ` to `R × R`. We'll have that the composite of two closed embeddings is a closed embedding, so this reduces the question to two easier questions: why is the inclusion a closed embedding and why is the map from `Rˣ` to `R × R` a closed embedding.

The inclusion from the kernel into `Rˣ` is just going to boil down to the statement that the kernel is closed, because it's bound to inherit the subspace topology. And the kernel is closed because `ringHaarChar` is continuous and {1} is closed because `Rˣ` is Hausdorff because `R` is Hausdorff. So that's the route you want to take.

As for why the map `Rˣ` to `R × R` is a closed embedding, this feels like it should be in the library already because it's not about things we defined like `ringHaarChar`, it's about things which were defined years ago. So look for it with `#loogle Units, Prod, TopologicalSpace` and we quickly find `Units.embedding_embedProduct` which says that it's an embedding. So all we have to do now is to prove that `Rˣ` is closed in `R × R`, and this isn't true for a general topological monoid but it is true for a Hausdorff one, so this is what you want to prove (and indeed I think I just PRed this recently so you might want to merge main and then find it.

All of these should be small lemmas with short proofs, which you can build up to give you the proof that M is compact.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+            obtain ⟨b, hb1, hb2⟩ := ha
+
+            sorry
+        · unfold incl₂3
+          simp only [Subgroup.subtype_apply, unitsEquivProdSubtype_apply_coe]
+          sorry
+    · exact Aux.C_compact K D
+  have h3 : (MtoQuot K D) '' (M K D) = Set.univ := by
+    ext x
+    refine ⟨by intro hx; simp, ?_ ⟩
```
kbuzzard, line 258: Clearly this is somehow just wasting time, because everything is a subset of univ, so there should be a lemma which saves you this time, and it seems to be `Set.eq_univ_iff_forall` (discovered with `rw?`). Looks like
```
  have h3 : (MtoQuot K D) '' (M K D) = Set.univ := by
    rw [Set.eq_univ_iff_forall]
    rintro ⟨a, ha⟩ 
    ...
```
is what you're looking for.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+  fun u => let p := (Units.embedProduct D_𝔸) (Subgroup.subtype (ringHaarChar_ker D_𝔸) u)
+           (p.1, MulOpposite.unop p.2)
+
+def incl₂ : ringHaarChar_ker D_𝔸 → Prod D_𝔸 D_𝔸 :=
+  (fun i => (i.1, i⁻¹.1)).comp (Subgroup.subtype (ringHaarChar_ker D_𝔸))
+
+def M : Set (ringHaarChar_ker D_𝔸) := Set.preimage (incl₂3 K D) (Aux.C K D)
+
+abbrev MtoQuot : (ringHaarChar_ker D_𝔸) → (ringHaarChar_ker D_𝔸 ⧸
+    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := Quot.mk _
```
kbuzzard, line 185: This is a bad move. There are at least two functions from G to G/H which are more idiomatic than the completely generic `Quot.mk` (which goes from X to X / ~). The first is `QuotientGroup.mk` and the second is simply the coercion `↑`. Introducing `Quot.mk` will force Lean to start explicitly writing the equivalence relation corresponding to "I am in the same coset as you" which will inevitably make things messier. The target of `Quot.mk` is `Quot r`, which is syntactically totally different to `G / H`.

Where you get stuck at the end of the proof your goal is of the form (ν : G / H) = (c * ν : G / H), except that you have exploded everything by using `Quot` and made the goal kind of unreadable. How about

```
abbrev MtoQuot (a : ringHaarChar_ker D_𝔸) : (ringHaarChar_ker D_𝔸 ⧸
    (MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype) := a
```
? Looking through the API for `QuotientGroup.mk` (just after the definition) I see `QuotientGroup.eq` which should be how to finish the proof.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+      obtain ⟨a, ha⟩ := x
+      obtain ⟨c, hc, ν, hν, eq, h31⟩ := Aux.antidiag_mem_C K D ha
+      simp only [Subgroup.comap_subtype, Set.mem_image, Subtype.exists]
+      refine ⟨ν, hν, h31, ?_ ⟩
+      simp_rw [MtoQuot, Subgroup.comap_subtype, eq]
+      rw [Quot.eq]
+      unfold QuotientGroup.leftRel
+
+
+      sorry
```
kbuzzard, line 269: Obviously at this point the goal has become unreadable, because you have been using `Quot` and unfolding things, which is definitely not what you want to do. If you had "kept control" in the proof then the goal at this point should be `↑⟨ν, hν⟩ = ↑⟨c * ν, ha⟩` where the up-arrow is the map from `G` to the space of cosets `G / H`. Recall here that `H` is the image of `incl` and you have a proof that `c` is in this image.

But now it looks to me like you're in trouble, because clearly the lemma you want to use next (I found it by reading the API for `QuotientGroup.mk` i.e. reading the statements of all the theorems just after the definition of this function in mathlib) is `QuotientGroup.eq`, which says `↑a = ↑b ↔ a⁻¹ * b ∈ H` and that inverse is on the wrong side for you. To put it another way, your goal is `νH = cνH` and what you want to happen next is that the c cancels into H but it doesn't because it's not `νcH`. Are you sure you've got the maths right? I'm not sure this goal is provable.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -169,8 +169,105 @@ lemma antidiag_mem_C {β : D_𝔸ˣ} (hβ : β ∈ ringHaarChar_ker D_𝔸) :
...
+        · unfold incl₂3
+          simp only [Subgroup.subtype_apply, unitsEquivProdSubtype_apply_coe]
+          sorry
+    · exact Aux.C_compact K D
+  have h3 : (MtoQuot K D) '' (M K D) = Set.univ := by
+    ext x
+    refine ⟨by intro hx; simp, ?_ ⟩
+    · intro hx
+      obtain ⟨a, ha⟩ := x
+      obtain ⟨c, hc, ν, hν, eq, h31⟩ := Aux.antidiag_mem_C K D ha
```
kbuzzard, line 261:
```suggestion
      obtain ⟨c, hc, ν, hν, rfl, h31⟩ := Aux.antidiag_mem_C K D ha
```
Always use the rfl trick, make things as simple as possible as quickly as possible.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -212,19 +309,175 @@ abbrev Dfx := (D ⊗[K] (FiniteAdeleRing (𝓞 K) K))ˣ
 noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
   Units.map Algebra.TensorProduct.includeLeftRingHom.toMonoidHom
 
+variable [MeasurableSpace (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)]
+  [BorelSpace (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)]
+
+def iso₁ : (D ⊗[K] NumberField.AdeleRing (𝓞 K) K)ˣ ≃*
+    Prod (D ⊗[K] NumberField.InfiniteAdeleRing K)ˣ (Dfx K D) := by
+  simp_rw [NumberField.AdeleRing, Dfx]
+  have h1 : D ⊗[K] (NumberField.InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K) ≃
```
kbuzzard, line 318: You'll never get anywhere with a `have` here: `have` is used for proofs and `≃` is data (it doesn't say "it's isomorphic to...", it says "the isomorphism is..." and `have` promptly forgets the isomorphism).

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -212,19 +309,175 @@ abbrev Dfx := (D ⊗[K] (FiniteAdeleRing (𝓞 K) K))ˣ
...
+    -- DirectSum.addEquivProd
+    -- TensorProduct.directSumRight
+
+    sorry
+  have := Units.mapEquiv h1
+  apply MulEquiv.prodUnits at this
+  sorry
+
+def rest₁ : ringHaarChar_ker D_𝔸 → Dfx K D :=
+  fun a => (iso₁ K D) a.val |>.2
```
kbuzzard, line 329: It's not clear to me that making definition after definition is a wise idea in general. Every definition clogs the system up. Sure abbreviate things if you're doing paper calculations, but every time you make a definition you should be making API for the definition, and here you make no API and immediately go on to make another definition. Definitions should be few and far between.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -212,19 +309,175 @@ abbrev Dfx := (D ⊗[K] (FiniteAdeleRing (𝓞 K) K))ˣ
...
+
 theorem NumberField.FiniteAdeleRing.DivisionAlgebra.units_cocompact :
     CompactSpace (Dfx K D ⧸ (incl₁ K D).range) := by
-  sorry
+  have h1 : Continuous (α K D) := by
+    unfold α rest₁
+    refine Continuous.comp' ?_ ?_
+    · exact { isOpen_preimage := fun s a ↦ a }
+    ·
+      -- this is certainly true by isomorphism and topology on RHS -- may need to explicitely show?
```
kbuzzard, line 346: No, this is not true (yet). If you `simp [iso₁]` the goal reduces to  `Continuous fun a ↦ ((sorry sorry) ↑(Quotient.out a)).2` so basically this function is not even defined, so you can't prove anything about it.

kbuzzard (commented):
I've reviewed about half of the PR, hopefully this is enough to keep you busy!

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: WIll do. I never intended for this to be merged anyway, its just to store my work and allow you to see where I am at. The maxHeartbeats were increased as simp wasnt finding anything and I wanted to test further - it will be removed when there is an actual PR.

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: Sure. When working through problems I try and prove everything together - which is maybe the wrong idea - then when cleaning for a PR I would separate them out.

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: I agree that I think something is wrong - we should be wanting to the right relation (as the idea of the proof is $$D^\times \backslash D^{(1)}$$ is compact - thus the $$c$$ should be on the right. At least that is if I am understanding and following everything correctly.

WilliamCoram: I think we should be wanting the right relation so really it should be $$H \nu =  H c \nu$$ - which would be fine.

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: When working through problems I always think "hmm, is my current goal actually something which should be an earlier lemma?" and if the answer is "yes" then I state the lemma earlier in the file, and then prove my goal using the earlier lemma, giving me a nice short proof, and then I start working on the earlier lemma instead. We want lots of nice short proofs not one long one, and my workflow generates them automatically. It also encourages me to find good names for my lemmas along the way (I don't want to write a short proof with `exact foo` in it because then I know it's still not actually finished; a short proof finishing with `exact [correct name of earlier lemma which I haven't actually proved yet]` is actually done and ready for merging.

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: Yes you're absolutely right, you're in trouble because *I* stated the lemma incorrectly! And it looks like you will indeed have to use `Quot`. But I would still recommend against unfolding things unless you temporarily want a look at how they've been defined. `rightRel` should have some API which will be what you need.

kbuzzard:
What is the status of this PR? Are you going to close it in favour of #617?

PR #556, opened 2025-05-24
Restricted product continuity

### LINKED ISSUE #530: A map on factors gives a map on restricted products
There should be a whole host of things all PRed to mathlib; an API for maps between restricted products (over the same index type) coming from maps on the factors. If the maps on the factors are group homs then the map on the restricted product is a group hom, if the maps on the factors are isomorphisms then the map on the restricted product is an isomorphism (this should come from id and comp lemmas), if the maps are continuous then the map on the restricted product is continuous etc etc, there are a whole host of things that we should be doing.

But let's stick to doing what we need for now, and this task is to show the first thing we need, namely that the map on the restricted product is well-defined. This is the sorry in `RestrictedProduct.congrRight` in `FLT/HaarMeasure/HaarChar/AddEquiv.lean`. This is just an exercise in filters really.

### LINKED ISSUE #531: Continuous functions on factors gives a continuous function on the restricted product
This is the continuity sorry in `Continuous.restrictedProductCongrRight` in the file `FLT/HaarMeasure/HaarChar/AddEquiv.lean`. I have sketched a proof using `RestrictedProduct.continuous_dom` in the LaTeX.

matthewjasper:
Proves a more general result which might be useful for proving continuity of base change for the finite adeles

Closes #530
Closes #531

### DIFF
--- a/FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
+++ b/FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
@@ -40,7 +40,7 @@ given maps on the factors. -/
 def congrRight (φ : (i : ι) → G i → H i)
     (hφ : ∀ᶠ i in ℱ, Set.MapsTo (φ i) (C i) (D i))
     (x : Πʳ i, [G i, C i]_[ℱ]) : (Πʳ i, [H i, D i]_[ℱ]) :=
-  RestrictedProduct.map G H id Filter.tendsto_id φ hφ x
+  map G H id Filter.tendsto_id φ hφ x
 
 end RestrictedProduct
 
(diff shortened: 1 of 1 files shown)


kbuzzard:
hmm. So I'm well aware that you can work in this generality (I guess I wrote `RestrictedProduct.map`) but I was kind of wondering whether it was "too general" and should be broken into two "steps", one where we map G_i to H_i and just have one index set I, and one where we map I to J and relate the restricted product over X_i for i in I to the restricted product indexed by j in J of (the product over i such that f(i)=j) of X_i). Do you think this is adding unnecessary noise and everything should be done with `map`? To be honest I am not completely clear on the advantage of separating out the map on the index sets and the map on the factors, but because in my Haar measure application the index sets weren't changing my instinct was to keep it simple. `MonoidHom.restrictedProductCongrRight` and `ContinuousMonoidHom.restrictedProductCongrRight` can also be generalized in this way, although my instinct is to leave the `Equiv`s alone.

matthewjasper:
I think that it's fine as is for now. For the `Equiv`s I think that splitting the equivalence makes some sense to make things easier to use/prove at first.

ADedecker:
I was not really active during the last few weeks, but I just want to point out https://github.com/leanprover-community/mathlib4/pull/24915/ which does the same proof (a bit shorter, but really the argument is the same AFAICT).

PR #561, opened 2025-05-27
feat: units of restricted products are (algebraically) restricted product of the units

### LINKED ISSUE #553: Units of a restricted product of rings (or monoids) is naturally isomorphic to restricted product of units
This is `MulEquiv.restrictedProductUnits` in `FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean`. It's 10 sorries, most of which should be trivial; I've filled in all the data. There's no LaTeX, hopefully it's a "follow your nose" job. Regarding the statements about filters, which I think are the only thing which isn't a one-liner, the point is that if x is a unit in a restricted product of M_i wrt submonoids A_i, then x^{-1} is also in this restricted product, so x_i is frequently in A_i, and x_i^{-1} is also frequently in A_i, so x_i and x_i^{-1} are frequently in the units of A_i. I think this is the hardest idea needed unless I missed something.

ADedecker:
Closes #553 (unless you expected the naturality lemmas ? I just filled in the sorry)


kbuzzard:
No you've done the job :-) Thanks!

PR #573, opened 2025-05-29
CI: don't build docs on non-`main` branches for real

YaelDillies:
The condition was wrong, as the workflow is triggered by push even when on non-`main` branches.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -117,7 +117,7 @@ jobs:
         run: scripts/build_docs.sh
 
       - name: Bundle dependencies
-        if: github.event_name == 'push'
+        if: github.ref == 'refs/heads/main'
         uses: ruby/setup-ruby@eaecf785f6a34567a6d97f686bbb7bccc1ac1e5c # v1.237.0
         with:
           working-directory: docs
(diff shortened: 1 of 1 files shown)


b-mehta (commented):
~~LGTM~~ ~~no~~ yes

pitmonticone:
It was right before the removal of the `on_push`-`on_pull_requests` distinction

b-mehta:
The current version should also be correct afaict

kbuzzard:
Thanks to all!

kbuzzard:
wait -- am I now doing everything twice?

Screenshot from 2025-05-31 13-17-35

kbuzzard:
yeah I seem to have two basically identical CIs now:

https://github.com/ImperialCollegeLondon/FLT/actions/runs/15363577891/job/43233670098?pr=580
https://github.com/ImperialCollegeLondon/FLT/actions/runs/15363578266/job/43233670713?pr=580

with different IDs and doing the same thing?

b-mehta:
> @b-mehta, this is what I was warning you about

Quite the opposite! The problem you showed me was caused by having two different `yml`, this one is caused by not restricting which branches CI runs on, which I warned you about :)

b-mehta:
The fix for this (which I mentioned on Thursday) is to make the `pull_request` CI only run on master, not on all branches.

kbuzzard:
What the heck? I just merged the huge PR #446 which was taking a really short time to pass CI (10 minutes) every time I pushed to the branch, and then I merged it into master and CI has passed and it's updated the blueprint and docs and it did it all in 21 minutes!

b-mehta:
what the hecke, even

b-mehta:
More seriously, that looks about right? Provided caching works (which it should most of the time), docs should take well under 10 minutes, and the blueprint build should also be fairly fast. Looking at older merges, this one https://github.com/ImperialCollegeLondon/FLT/actions/runs/14973502361 from two weeks ago also took 21 minutes, so I'm not really surprised by the new one.

kbuzzard:
I had thought that "given that CI sometimes used to take 50 minutes, it will always take 50 minutes when I merge something"

b-mehta:
Yes, it took 21 minutes when you merged things two weeks ago, and it takes 21 minutes when you merge things these days! The point of this PR was to impact what happens in PRs, not what happens when you merge to main

PR #575, opened 2025-05-30
feat: ringHaarChar_continuous 

### LINKED ISSUE #516: The Haar character of a ring is continuous
This is a little delicate. There's a proof in the LaTeX -- see [here](https://imperialcollegelondon.github.io/FLT/blueprint/Haar_char_project.html#MeasureTheory.ringHaarChar_continuous). It's taken straight from Gouezel's proof on the Zulip. This is `ringHaarChar_continuous` in `FLT/HaarMeasure/HaarChar/Ring.lean`.

DavidLedvinka:
Will close #516 .

The technical part of the proof is done but the finishing step in the comments references a lemma "`mulEquivHaarChar_mul_integral` in the AddEquiv file" which I believe doesn't exist. Alternatively the lemma referenced in the blueprint `ringHaarChar_mul_integral` comes after this lemma and can't be put before with the way things are organized. We need a version of this lemma that comes before `ringHaarChar_continuous` to finish. I could just inline this part of the proof but I figured I should ask first!


kbuzzard:
Draft PRs cannot be merged but I'm assuming from your comment that it's ready for review? Can you un-draft it?

DavidLedvinka:
Sorry yes I was trying to bring attention to the question! To be precise I need Lemma 9.11 in the blueprint. This is a consequence of `addEquivAddHaarChar_smul_integral_map` as mentioned in the blueprint so I could just inline that step directly.

PR #589, opened 2025-05-31
chore: remove proof that flat <-> tfree for PID

kbuzzard:
This is being upstreamed to mathlib in https://github.com/leanprover-community/mathlib4/pull/25334 . When this is merged we can finally remove the FLT/HIMExperiments directory.


kbuzzard:
oops git mess

PR #591, opened 2025-06-01
feat: fill in two sorries in RestrictedProduct

### LINKED ISSUE #569: Canonical map from finite product of restricted products to restricted product of finite products
This is two easy sorries. I just couldn't figure out how to get `filter_upwards` to do them so I've left them as exercises. It's the statement "intersection of finitely many elements in the cofinite filter is cofinite" heavily disguised. It's the sorries in `Equiv.restrictedProductPi` and ` Equiv.restrictedProductMatrix` in `FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean`. I wonder whether it would have been easier to fill them in than write this issue :-/ Let me know if you can solve them with `filter_upwards`!

b-mehta:
null

Closes #569

### DIFF
--- a/FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
+++ b/FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
@@ -23,6 +23,8 @@ lemma mul_apply {S : ι → Type*} [(i : ι) → SetLike (S i) (R i)] {B : (i : 
     [(i : ι) → Mul (R i)] [∀ (i : ι), MulMemClass (S i) (R i)]
     (x y : Πʳ (i : ι), [R i, ↑(B i)]_[ℱ]) (i : ι) : (x * y) i = x i * y i := rfl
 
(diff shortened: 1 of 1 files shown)


b-mehta:
awaiting-review

FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
```diff
@@ -205,17 +207,17 @@ def Homeomorph.restrictedProductProd {ι : Type*}
       continuous_invFun := sorry -- FLT#568
 
 -- Is there a mathlibism for {f | ∀ j, f j ∈ C j i}?
+-- Yes: Set.pi Set.univ
```
kbuzzard, line 210: So can you change `Equiv.restrictedProductPi` to use this, or is there a catch?

b-mehta: The catch is defeqs: Set.pi Set.univ would give `{f | ∀ j ∈ univ, f j ∈ C j i}` which `simp`s to what you want, but isn't exactly the same. imo we should keep this as is since it's easier to use, but it doesn't make a big difference either way

kbuzzard:
Thanks! I'm going to clarify the comment.

PR #592, opened 2025-06-01
feat: QuotientGroup.mk_image_finite_of_compact_of_open

### LINKED ISSUE #563: If U and V are compact open then UgV/V is finite
This is lemma `QuotientGroup.mk_image_finite_of_compact_of_open` in file `FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Concrete.lean`,
and [https://imperialcollegelondon.github.io/FLT/blueprint/HeckeOperator_project.html#QuotientGroup.mk_image_finite_of_compact_of_open](https://imperialcollegelondon.github.io/FLT/blueprint/HeckeOperator_project.html#QuotientGroup.mk_image_finite_of_compact_of_open) in the blueprint.

The statement is this. Let G be a topological group and let U,V be compact open subgroups. Even though we don't assume V is normal, there's still a quotient space G / V, a topological space representing the left cosets of V in G. Now say g is in G. I claim that the image of the double coset space UgV := {ugv | u in U, v in V} in G/V is finite (or equivalently that `UgV` is a finite union of single cosets g_iV), and the proof is that UgV/V is compact (because it's a continuous image of U x V) and discrete (because G/V has the discrete topology, as V is open so the preimage of any subset of G/V in G is open as it's a union of cosets of V).

This should be doable in one sitting for anyone who knows their way around the topology part of the library.

DavidLedvinka:
null

Closes #563

### DIFF
--- a/FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Concrete.lean
+++ b/FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Concrete.lean
@@ -52,12 +52,11 @@ via a topological argument, which we abstract here.
 
 section finiteness
 
-open Set
+open Topology Set
 
 variable {G : Type*} [Group G] [TopologicalSpace G] [IsTopologicalGroup G]
     {g : G} {U V : Subgroup G}
 
-open Topology in
 open scoped Pointwise in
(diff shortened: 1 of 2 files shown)


DavidLedvinka:
awaiting-review

PR #593, opened 2025-06-01
restricted and binary products

b-mehta:


FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
```diff
@@ -176,46 +184,90 @@ def MulEquiv.restrictedProductUnits {ι : Type*} {ℱ : Filter ι}
...
+    intro S hS
+    rw [Equiv.invFun_as_coe, Equiv.restrictedProductProd_symm_comp_inclusion]
+    fun_prop
+
+end binary
+
+section pi
 
 -- Is there a mathlibism for {f | ∀ j, f j ∈ C j i}?
+-- Yes: Set.pi Set.univ
```
kbuzzard: So can we use it in the next def?

b-mehta:
awaiting-review

FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
b-mehta: (discussed in #591)

kbuzzard:
You didn't clarify the comment, so I've clarified the comment and now there are two merge conflicts on this one (the comment and something else)

b-mehta:
Fixed: both sides were defeq, but one used the new explicit constructor whereas the other used an anonymous one. Nothing mathematically interesting there :)

b-mehta:
I don't know if you want to change anything based off this, but for both issue 569 and 570 I formalised a different proof than what you sketched. For 569 I showed that the equivalence is functorial (in the filter) where the nat trans is continuous, and that the equivalence for the principal filter is a homeo. Then passing through the nat trans, we get that the equivalence for the cofinite filter (ie the usual case) is also a homeo

FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
```diff
@@ -178,33 +184,78 @@ def MulEquiv.restrictedProductUnits {ι : Type*} {ℱ : Filter ι}
...
-    {C : (i : ι) → Set (A i)}
-    {D : (i : ι) → Set (B i)} :
+@[simps]
+def Equiv.restrictedProductProd :
     Πʳ i, [A i × B i, C i ×ˢ D i]_[ℱ] ≃ (Πʳ i, [A i, C i]_[ℱ]) × (Πʳ i, [B i, D i]_[ℱ]) where
-      toFun x := (⟨fun i ↦ (x i).1, by filter_upwards [x.2] with i using And.left⟩,
-                  ⟨fun i ↦ (x i).2, by filter_upwards [x.2] with i using And.right⟩)
-      invFun yz := ⟨fun i ↦ (yz.1 i, yz.2 i), by
+      toFun x := (congrRight (fun i (t : A i × B i) ↦ t.1) (by simp +contextual [Set.MapsTo]) x,
+                  congrRight (fun i (t : A i × B i) ↦ t.2) (by simp +contextual [Set.MapsTo]) x)
```
b-mehta, line 201: Just to record it, this is defeq to the old map, but this one can be proved continuous by fun_prop (for any filter!)

PR #594, opened 2025-06-01
feat: restricted products and finite products

### LINKED ISSUE #570: Restricted products (wrt opens) commute with finite products
The claim is that restricted products commute with finite products. This is vacuous if the product is empty and it's `Homeomorph.restrictedProductProd` for a binary product, so now it should be mathematically straightforward to deduce it for finite products. The blueprint just says "induction on the size of the finite set". The declaration is `Homeomorph.restrictedProductPi` in `FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean`.

b-mehta:
Depends on #593 and #591

Also proves the finitary generalisation for the characterisation of continuous functions out of a product of restricted products, which should make the main result easier

Avoids induction, since that's messy.

Closes #570


kbuzzard:
I should say that the only reason I wanted this is for the analogous result for matrices (matrix m n (restricted prod) homeo to restricted prod of (matrix m n factor)) and it wasn't even clear to me that the finite product result would be easily applicable. One would either need to curry (matrix m n X is m -> n -> X not (m x n) -> X) and then check that everything commutes with curry, or do matrix m n (restricted prod) is m -> (n -> restricted prod) which is homeo to m -> (restricted prod of (n -> factor)) which is homeo to restricted prod of (m -> n -> factor).

b-mehta:
> or do matrix m n (restricted prod) is m -> (n -> restricted prod) which is homeo to m -> (restricted prod of (n -> factor)) which is homeo to restricted prod of (m -> n -> factor).

I prefer this idea! Plus the new lemma added in this PR is something that's useful for mathlib more broadly, outside of matrix things

kbuzzard:
can you merge and fix conflicts?

b-mehta:
awaiting-review

FLT/Mathlib/Topology/Algebra/RestrictedProduct.lean
```diff
@@ -267,32 +278,76 @@ def Equiv.restrictedProductPi {ι : Type*} {ℱ : Filter ι} {n : Type*} [Fintyp
     {A : n → ι → Type*}
     {C : (j : n) → (i : ι) → Set (A j i)} :
     Πʳ i, [Π j, A j i, {f | ∀ j, f j ∈ C j i}]_[ℱ] ≃ Π j, Πʳ i, [A j i, C j i]_[ℱ] where
-      toFun x j := .mk (fun i ↦ x i j) (by filter_upwards [x.2] with i h using h j)
-      invFun y := .mk (fun i j ↦ y j i) (by simp)
-      left_inv x := by ext; rfl
-      right_inv y := by ext; rfl
+  toFun x j := congrRight (fun i t ↦ t _) (by simp +contextual [Set.MapsTo]) x
```
kbuzzard, line 281: If you're adding those variables, why not remove them from the def of `Equiv.restrictedProductPi`?

kbuzzard: I'll remove them myself

PR #597, opened 2025-06-01
feat: ContinuousMulEquiv.piUnits

Ruben-VandeVelde:
Fixes #581.

Closes #581

PR #599, opened 2025-06-01
fix: mark ContinuousMulEquiv.piUnits as done

Ruben-VandeVelde:
See  #597.


kbuzzard:
oops thanks. 

I think I need CI to check that everyone who claims to close a task also edits the blueprint :-) Or we need to persuade Patrick to make it happen automatically!

PR #604, opened 2025-06-01
Proposal for FLT task521

AbdulShabazz:
The proof strategy reduces the problem to showing that the Haar characteristic is multiplicative with respect to products, then uses the special structure of finite types (built from `Empty` and `Option`) to complete the induction.

The proof's key components:

1. **Structure**: The proof uses `Fintype.induction_empty_option` which is the appropriate induction principle for finite types in Lean 4.

2. **Base case** (empty type):
   - Identifies `Π i : Empty, H i ≃ₜ* Unit`
   - Shows that `piCongrRight ψ` is essentially the identity on Unit
   - Uses `mulEquivHaarChar_eq_one_of_compactSpace` to conclude the Haar characteristic is 1
   - Both sides equal 1 (empty product is 1)

3. **Inductive step** (Option ι):
   - Uses the isomorphism `(Π i : Option ι, H i) ≃ₜ* H none × (Π i : ι, H (some i))`
   - Shows `piCongrRight` commutes with this decomposition
   - Applies `mulEquivHaarChar_prodCongr` to get the product formula
   - Uses the induction hypothesis `IH`
   - Applies `Fintype.prod_option` to rearrange the finite product

4. **Missing pieces that might need attention**:
   - The proof assumes `mulEquivHaarChar_prodCongr` exists (which is referenced in the file as `FLT#520`)
   - The proof assumes various lemmas about `MulEquiv.piOptionEquivProd` and `MulEquiv.piEmpty` exist in Mathlib

The `end pi` at the bottom closes the section that was opened earlier in the file.

This proof should work once the dependency `mulEquivHaarChar_prodCongr` is also proven!


FLT/HaarMeasure/HaarChar/AddEquiv.lean
```diff
@@ -270,12 +270,122 @@ variable {ι : Type*} {H : ι → Type*} [Π i, Group (H i)] [Π i, TopologicalS
     [∀ i, IsTopologicalGroup (H i)] [∀ i, LocallyCompactSpace (H i)]
     [∀ i, MeasurableSpace (H i)] [∀ i, BorelSpace (H i)]
 
+
+set_option maxHeartbeats 2000000000
```
kbuzzard: I am not sure that this is acceptable. Why do you need it?

lakefile.lean
```diff
@@ -0,0 +1,32 @@
+
+import Lake
```
kbuzzard: You definitely should not be committing a system file in this PR -- we have all the system files we need for the project.

FLT/HaarMeasure/HaarChar/AddEquiv.lean
AbdulShabazz: LEAN elaboration/inference timeouts after a depth of 2000000. Adjusting maxHeartbeats was an ad-hoc fix.

lakefile.lean
AbdulShabazz: Ah. lakefile.toml files require Mathlib recompilation (and are deprecated in LEAN 4) upon reloading the project, which is a bit of an aggravation. A .lean file caches Mathlib to /User/lean/.olean. I removed this personal convenience in a subsequent commit.

AbdulShabazz:
@js2357, excellent! Can you please provide a Propose #number/PR #number and or a branch name with the filename.lean, if you are originating the proof in LEAN?

AbdulShabazz:
Dude... you are being such a troll: I requested the PR **before** the merge! Check the timestamps. ...I don't like you.

AbdulShabazz:
Leave it to the pros. Not straightforward enough for me.

PR #609, opened 2025-06-03
feat: Subgroup Lemmas 

DavidLedvinka:


DavidLedvinka:
awaiting-review

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -34,6 +34,26 @@ lemma _root_.MeasurableEmbedding.IsMulLeftInvariant_comap {G H : Type*}
     rw [this, ← map_apply (by fun_prop), IsMulLeftInvariant.map_mul_left_eq_self]
     exact hφ.measurableSet_image.mpr hs
 
+@[to_additive]
+lemma _root_.MeasurableEmbedding.IsMulRightInvariant_comap {G H : Type*}
```
YaelDillies:
```suggestion
lemma _root_.MeasurableEmbedding.isMulRightInvariant_comap {G H : Type*}
```

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -34,6 +34,26 @@ lemma _root_.MeasurableEmbedding.IsMulLeftInvariant_comap {G H : Type*}
...
+    [Group G] [MeasurableSpace G] [MeasurableMul G]
+    [Group H] [MeasurableSpace H] [MeasurableMul H]
```
YaelDillies: Can you get away with
```suggestion
    [Monoid H] [MeasurableSpace H] [MeasurableMul H]
```
or maybe `Monoid G` instead if ever I got the side wrong?

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -47,24 +67,27 @@ variable {G α : Type*} [Group G] [MeasurableSpace G] [MeasurableSpace α]
...
+lemma isMulLeftInvariant_subtype_coe (μ : Measure G) [μ.IsMulLeftInvariant]
+  (hH : MeasurableSet (H : Set G)) : (μ.comap Subtype.val : Measure H).IsMulLeftInvariant :=
+  have hφ : MeasurableEmbedding H.subtype := MeasurableEmbedding.subtype_coe hH
+  hφ.IsMulLeftInvariant_comap μ
 
 @[to_additive]
-instance (μ : Measure G) [μ.IsMulRightInvariant] :
-    (μ.comap Subtype.val : Measure H).IsMulRightInvariant where
-  map_mul_right_eq_self g := sorry -- https://github.com/ImperialCollegeLondon/FLT/issues/276
+lemma isMulRightInvariant_subtype_coe (μ : Measure G) [μ.IsMulRightInvariant]
```
YaelDillies:
```suggestion
lemma isMulRightInvariant_subtypeVal (μ : Measure G) [μ.IsMulRightInvariant]
```

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -47,24 +67,27 @@ variable {G α : Type*} [Group G] [MeasurableSpace G] [MeasurableSpace α]
...
 instance : MeasurableMul H where
   measurable_mul_const c := by measurability
-  measurable_const_mul c := sorry -- https://github.com/ImperialCollegeLondon/FLT/issues/276
+  measurable_const_mul c := Measurable.subtype_mk (by measurability)
 
 @[to_additive]
-instance (μ : Measure G) [μ.IsMulLeftInvariant] :
-    (μ.comap Subtype.val : Measure H).IsMulLeftInvariant where
-  map_mul_left_eq_self g := sorry -- https://github.com/ImperialCollegeLondon/FLT/issues/276
+lemma isMulLeftInvariant_subtype_coe (μ : Measure G) [μ.IsMulLeftInvariant]
```
YaelDillies:
```suggestion
lemma isMulLeftInvariant_subtypeVal (μ : Measure G) [μ.IsMulLeftInvariant]
```

YaelDillies:
awaiting-author

FLT/Mathlib/MeasureTheory/Group/Action.lean
DavidLedvinka: This question actually made me realize IsMulLeftInvariant in mathlib is defined kind of weirdly. Its defined as `(∀ g : G, map (g + ·) μ = μ` which means for any g ∈ G and measurable set S , 
$μ ( ( ⋅ ∗ g )^{-1} S ) = μ ( S )$ (not what you would expect which is $μ ( g S ) = μ ( S )$. This is fine for groups but is not good if you want it to work on monoids. For example the counting measure on $\mathbb{N}$ is not mul left invariant and the injection $\mathbb{N} \to \mathbb{Z}$ is a counter example to this theorem!

DavidLedvinka: `Monoid H` however is doable with this definition. Will make the upgrade tomorrow.

DavidLedvinka: Ok after some research it does seem like this definition is standard. Invariance under translation here can be interpreted as invariant under the pushforward of the translation operator (which pulls back sets). Unfortunately this is not equivalent to invariance of the measure of translation of sets (for monoids).

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -14,10 +14,12 @@ section MeasurableEmbeddingComap
 
 open MeasureTheory Measure
 
+example {a b c : ℝ} (h : a = b) : c * a = c * b := by exact congrArg (HMul.hMul c) h
+
```
YaelDillies:
```suggestion
```

FLT/Mathlib/MeasureTheory/Group/Action.lean
```diff
@@ -15,9 +15,9 @@ section MeasurableEmbeddingComap
 open MeasureTheory Measure
 
 @[to_additive]
-lemma _root_.MeasurableEmbedding.IsMulLeftInvariant_comap {G H : Type*}
+lemma _root_.MeasurableEmbedding.isMulLeftInvariant_comap {G H : Type*}
```
kbuzzard, line 18: Could this be an `instance`? Oh I see -- it can't be because MeasurableEmbedding isn't a class! Rotten luck!

PR #610, opened 2025-06-03
feat: suppress sorry warnings at `lake build`

digama0:
See also leanprover/lean4#8611 and [#lean4 > Silent sorry @ 💬](https://leanprover.zulipchat.com/#narrow/channel/270676-lean4/topic/Silent.20sorry/near/503537964). A more permanent version will be PR'd to mathlib. (And if the FRO addresses leanprover/lean4#8611 then this won't be needed in the first place.)


kbuzzard:
closed in favour of commit 9feced6 which adds `warn.sorry = false` to `lakefile.toml`.

PR #612, opened 2025-06-04
chore: bump mathlib

grunweg:
#605 should be merged first. Also, I haven't checked if any further results in `RestrictedProducts` can now be removed.


FLT/GlobalLanglandsConjectures/GLzero.lean
```diff
@@ -67,9 +67,7 @@ def ofComplex (c : ℂ) : AutomorphicFormForGLnOverQ 0 ρ := {
     is_slowly_increasing x := ⟨‖c‖, 0, by simp⟩
     is_finite_cod := by
       intros x
-      rw [FiniteDimensional]
-      rw [annihilator]
-      simp only [id_eq, eq_mpr_eq_cast, cast_eq]
```
grunweg, line 72: This simp call no longer makes any progress: I have removed it.

kbuzzard:
@grunweg I believe you should also have removed `rev = "v4.20.0"` from the mathlib entry in this project's `lakefile.toml` with this PR. In fact I am surprised you managed to bump at all with this in the toml (I discovered this when I attempted to bump again today).

PR #613, opened 2025-06-04
chore: enable precisely mathlib's linters

grunweg:
Use the new mathlib linter set to enable all of mathlib's syntax linters. This change sped up mathlib unexpectedly, so this could be nice to have as well.
Depends on #612.


kbuzzard:
I'm just awaiting a green tick on master for this one (we are rebuilding docs after the bumps; this always takes an age)

PR #614, opened 2025-06-04
feat(blueprint.yml): run shake in CI

grunweg:
@pitmonticone I'm interested if you have any comments here. (No worries, this is not urgent.)

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -77,6 +77,13 @@ jobs:
       - name: Lint project
         run: env LEAN_ABORT_ON_PANIC=1 ~/.elan/bin/lake exe runLinter FLT
 
+      - name: check for unused imports
+        id: shake
+        uses: leanprover-community/gh-problem-matcher-wrap@20007cb926a46aa324653a387363b52f07709845 # 2025-04-23
+        with:
+          linters: gcc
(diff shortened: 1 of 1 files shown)


kbuzzard:
Yeah I've been running shake manually every week or two -- this would be welcome, if you can figure out how to make it work!

azide0x37:
claim 😉

https://github.com/azide0x37/FLT/pull/1/files

with.run doesn't use the shell, invoking specifically works.

Could also add step ahead of lake invocations:

```
- name: Make lake available on PATH
  run: echo "$HOME/.elan/bin" >> "$GITHUB_PATH"
```

Which allows lake syntax below when using `env`: 
```
- name: check for unused imports
  id: shake
  uses: leanprover-community/gh-problem-matcher-wrap@20007cb926a46aa324653a387363b52f07709845
  env:
    LEAN_ABORT_ON_PANIC: "1"
  with:
    linters: gcc
    run: lake exe shake --gh-style
```

kbuzzard:
This is pre-module system

grunweg:
(And the new shake tool is not fully ready for projects of this scale yet: at the moment, it's not even enabled in mathlib mathlib yet. Waiting for that to happen certainly makes sense!)

PR #623, opened 2025-06-09
Cache mathlib builds

AbdulShabazz:
I cache Mathlib builds, until invalidation. This reduces Mathlib library build lengths from an hour to only a few minutes,for updated files.

I also perform a `lake update`, followed by `lake build`. You may want to cherry pick this commit.

Regards,
Abdul S.


lakefile.toml
```diff
@@ -21,3 +21,6 @@ git = "https://github.com/PatrickMassot/checkdecls.git"
 [[lean_lib]]
 name = "FLT"
 globs = ["FLT", "FermatsLastTheorem"]
+
+[[dependancies.FLT]]
```
grunweg, line 25: Are you sure this works, given the typo-ed "dependencies"? (I don't know if this is an actual section of `lakefile`s.)

AbdulShabazz: Oops! Perhaps it's cartel malware trolling my repo and or my home machine. It originally didn't work for me, and then with this change, it did work for me. I have no control group to measure against. I employ formal software development principles which are largely created by cartels for guaranteed loopholes in production code, such as Exception-handling, which guarantees a fuzzable landing address if hackers can get a portion of precompiled code to crash. I f*** hate malware and malware developers; I wish law enforcement would cease to adopt their tooling and just murder them.

PR #624, opened 2025-06-09
feat: prove `group_theory_lemma`

wupr:
I promised Kevin I'd open a PR on this over a year ago and I'm finally doing it.

I have a sorry-free proof (~400 lines) of `group_theory_lemma` that I'll tidy up and push here in the next couple of weeks.


kbuzzard:
Note that your fix is still wrong -- I have corrected the statement in `main`. AlphaProof found the error!

PR #626, opened 2025-06-10
modified:   lakefile.toml (Cache Mathlib builds until invalidation)

AbdulShabazz:
Cache Mathlib builds, until invalidation. This reduces Mathlib incremental builds from a span of hours, to only a few minutes.


AbdulShabazz:
This configuration should be an option, perhaps in the documentation.

PR #634, opened 2025-06-12
chore: get project linting again

kbuzzard:
I think that more mathlib linters means more linter problems? We fix them here.


kbuzzard:
Hmm I'm now totally confused. I was expecting to see all the warnings which I saw in https://github.com/ImperialCollegeLondon/FLT/actions/runs/15622452709/job/44010158069?pr=632

kbuzzard:
huh, I had to build manually using Damiano's hack
```
lake build FLT.lean | grep "^\(warning:\|error:\|info:\|Build completed\)" | grep -v "^warning: .*: declaration uses 'sorry'"
```
to see the other warnings (other than the complaints about `sorry`). They're all now dealt with.

PR #640, opened 2025-06-16
ci: Bump update action.

Vierkantor:
This new version does not error if no upgrades exist, and has a 'stable'option for skipping prereleases and Mathlib `master` commits.

Tested on https://github.com/Vierkantor/lean-update-tester.


pitmonticone (approved):
Thanks!

PR #646, opened 2025-06-23
Fujisaki: finiteDoubleCoset proof

WilliamCoram:
Full proof now done. 
To do: Move generalised lemmas to mathllib


WilliamCoram:
awaiting-review

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -241,7 +242,9 @@ end NumberField.AdeleRing.DivisionAlgebra
 
 section FiniteAdeleRing
 
-variable [FiniteDimensional K D]
+/- removing this variable for now so that CI works; will need to be readded for the proof of
+   units_cocompact -/
+-- variable [FiniteDimensional K D]
```
kbuzzard: This definitely should be removed because it's already there, it's on line 37 so it's in place throughout the file.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
   sorry
 
+lemma Doset.mem_quotToDoset_iff {G : Type*} [Group G] (H K : Subgroup G)
+    (i : Quotient (H : Set G) K) (g : G) : g ∈ quotToDoset H K i ↔ mk H K g = i := by
+  constructor
+  · intro hg
+    simp_rw [mk_eq_of_doset_eq (doset_eq_of_mem hg), Quotient.out_eq]
+  · intro hg
+    rw [← out_eq' _ _ i] at hg
+    exact mem_doset.mpr ((eq _ _ _ g).mp hg.symm)
```
kbuzzard: All the general lemmas about double cosets should definitely be in `FLT/Mathlib/GroupTheory/DoubleCoset.lean` rather than this file (and ideally PRed to mathlib as well :-) but that's another story, and nothing to do with this PR)

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+    obtain ⟨_, ⟨i, _, rfl⟩, T, ⟨hi, rfl⟩, hT : g ∈ quotToDoset H K i⟩ := hI g
+    rw [Doset.mem_quotToDoset_iff] at hT
+    simpa [← hT] using hi
+  · intro _
+    cases nonempty_fintype (Quotient (H : Set G) K)
+    use Finset.univ
+    simpa using Doset.union_quotToDoset H K
+
+-- dosets of elements of the double cosets H\G/K descends to a cover of H\G
+-- no idea what this should be called
```
kbuzzard: If you don't know what to call something, and it doesn't really look like an analogue of anything which is already there, then you can do a lot worse than just starting to write the "mathlib code" for the concepts in the statement (until you figure that you've written enough). So here for example `Doset.union_image_mk_rightRel` would be a name which would probably get through.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+  obtain ⟨y, hy⟩ := Quot.exists_rep x
+  have ⟨i, hi⟩ : ∃ i : Doset.Quotient H K,
+      y ∈ Doset.doset (Quotient.out i) H K  := by
+    contrapose Cover_Dfx
+    refine (Set.ne_univ_iff_exists_notMem (⋃ q, quotToDoset H K q)).mpr ?_
+    exact ⟨y, by simpa using Cover_Dfx⟩
+  exact ⟨i, y, hi, hy⟩
+
+-- no idea what to name this
+-- If U is open, all images of dosets of elements of H\G/K into H\G are open
```
kbuzzard: This has topology in so doesn't necessarily go in that mathlib PR. But it seems to me that this lemma/proof is misguided. What we should really be doing is

a) the quotient map G -> H \ G is an `OpenQuotientMap`. This should be a lemma; the proof should follow `QuotientGroup.isOpenQuotientMap_mk`, which proves the same for `G / H`.
b) If U is open then any double coset HgU is open (because it's a union of open sets). This should be a lemma if it's not there already.
c) so by (a) the image of this double coset in H \ G is open (and now probably we don't even need to state the lemma because it follows immediately from (a) and (b)). 

I feel like you're proving a special case of (a) somewhere in the lemma but it should be extracted.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+theorem Doset.leftcoset_isOpen {G : Type*} [Group G] [TopologicalSpace G] [ContinuousConstSMul G G]
```
kbuzzard:
```suggestion
theorem Doset.leftcoset_isOpen {G : Type*} [Group G] [TopologicalSpace G] [ContinuousMul G]
```
`ContinuousConstSMul` is for things acting on other things. If `a b : G` then the way to get a acting on b is `a * b` not `a \smul b` so we should be using `Mul` not `SMul` here.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+      refine ⟨a', by simpa using ha'⟩
+    refine ⟨⟨ a⁻¹, by simp only [inv_mem_iff, SetLike.coe_mem]⟩, eq_inv_mul_of_mul_eq ha⟩
+  refine ⟨a.1, ?_⟩
+  simp only [Subtype.coe_prop, SetLike.mem_coe, true_and]
+  exact ⟨1, Subgroup.one_mem K, by simpa using ha⟩
+
+
+/-- CI. -/
+local instance : SMul (Dfx K D) (Dfx K D) where
+  smul := HMul.hMul
```
kbuzzard: This shouldn't be there, my guess is that it's only needed because you introduced `SMul` earlier instead of using `Mul`.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+
+open scoped TensorProduct.RightActions
+local instance : ContinuousConstSMul (Dfx K D) (Dfx K D) where
+  continuous_const_smul a := by
+    simp only [smul_eq_mul]
+    exact continuous_mul_left a
```
kbuzzard: This shouldn't be needed either.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -263,9 +266,122 @@ noncomputable abbrev incl₁ : Dˣ →* Dfx K D :=
...
+  refine ⟨a.1, ?_⟩
+  simp only [Subtype.coe_prop, SetLike.mem_coe, true_and]
+  exact ⟨1, Subgroup.one_mem K, by simpa using ha⟩
+
+
+/-- CI. -/
+local instance : SMul (Dfx K D) (Dfx K D) where
+  smul := HMul.hMul
+
+open scoped TensorProduct.RightActions
```
kbuzzard: This opens the scope for the entire rest of the file. Is that what you meant to do?

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: Thanks for pointing this out!

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: It was meant to have ``open ... in``, but the need for this instance has been removed by the changing of ``smul`` to ``mul``.

kbuzzard:
This has conflicts.

awaiting-author

WilliamCoram:
awaiting-review

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -3,6 +3,7 @@ Copyright (c) 2024 Kevin Buzzard. All rights reserved.
 Released under Apache 2.0 license as described in the file LICENSE.
 Authors: Kevin Buzzard, William Coram
 -/
+
```
kbuzzard: This is not the correct style. Can you delete the line? Maybe you should glance through your diffs after making a PR and ask yourself "did I mean to PR every change I made?"

FLT/Mathlib/GroupTheory/DoubleCoset.lean
```diff
@@ -0,0 +1,85 @@
...
+  have ⟨i, hi⟩ : ∃ i : DoubleCoset.Quotient H K,
+      y ∈ doubleCoset (Quotient.out i) H K  := by
+    contrapose Cover_Dfx
+    refine (Set.ne_univ_iff_exists_notMem (⋃ q, quotToDoubleCoset H K q)).mpr ?_
+    exact ⟨y, by simpa using Cover_Dfx⟩
+  exact ⟨i, y, hi, hy⟩
+
+theorem DoubleCoset.isOpen_doubleCoset {G : Type*} [Group G] [tG : TopologicalSpace G]
+    [ContinuousMul G] (H K : Subgroup G) (hK : IsOpen (K : Set G)) (i : DoubleCoset.Quotient H K) :
+    @IsOpen G tG (doubleCoset (Quotient.out i) H K) := by
```
kbuzzard: Do you really need the `@`?

WilliamCoram:
> Yeah this is what this PR should look like -- thanks a lot! You should consider PRing the double coset stuff to mathlib to see what they think of it, but it's fine for FLT. Can you just deal with my two nits and then I'll merge.

I already have a [PR](https://github.com/leanprover-community/mathlib4/pull/27229) for some of the results.

FLT/Mathlib/GroupTheory/DoubleCoset.lean
WilliamCoram: I have changed it to ``IsOpen (X := G) (...)``. It cannot infer the topological space without the (X := G), and I am unsure what is preferred.

WilliamCoram:
awaiting-review

FLT/Mathlib/GroupTheory/DoubleCoset.lean
kbuzzard: Huh, I guess the reason is that `H : Subgroup G` rather than `H : Set G`.

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: (you can just resolve nits like this when they're dealt with)

kbuzzard:
Thanks!

Issue #521, opened 2025-05-18
Fudge factor for finite product of isomorphisms is product of fudge factors


kbuzzard:
This is `MeasureTheory.mulEquivHaarChar_piCongrRight` in `FLT/HaarMeasure/HaarChar/AddEquiv.lean`. This should be a straightforward induction once you find the right induction principle,
with the key step being the previous lemma `MeasureTheory.mulEquivHaarChar_prodCongr` (which does two factors). If you want to do finsets first, that's fine.

AbdulShabazz:
Claim #521

js2357:
@AbdulShabazz I'm not actually sure whether I can claim it yet because you didn't disclaim it correctly. (Notice that it never left the "In Progress" column on the dashboard.)

AbdulShabazz:
Okay, hold on. I'll revisit the contribution guidelines...

AbdulShabazz:
Withdraw #604

PR #653, opened 2025-06-30
Update Example.lean

roozbeh-mohit:
Hello, I just made a tiny addition to one of the lemmas to get some feedback and possibly contribute to the project in the future. Is what I have added for the proof of "lemma injective_hRat :" helpful? Thanks!


AbdulShabazz:
Hello, Roozbeh! And thanks much for the contribution(s)!! If the changes you propose directly indeed relate to the proof of Fermat's Last Theorem (FLT) and pass CI, of course I would think they are welcome; however if they fail CI, they will likely have to be refined. For example, I've found discovered a proof `mulEquivHaarChar_piCongrRight_compactSpacs` which relates to task #521, however it is not in the general case, and thus does not apply to FLT; so I will have to wait for a proof on the general case to submit, or perhaps open a direct PR to Mathlib...

Regards,
Abdul S.

kbuzzard:
@roozbeh-yz Thanks! It's not clear to me that your proposed proof is going in the right direction; my vision is that injectivity will just be proved directly from flatness of Zhat.

Issue #584, opened 2025-05-31
Hecke operators U_alpha and U_beta at a bad prime commute


kbuzzard:
If U looks like (a *;0 a) at a place v then we define the Hecke operator U_{v,alpha} for alpha nonzero in O_v to be the Hecke operator associated to the matrix (alpha 0; 0 1). The claim is that U_{v,alpha} U_{v,beta} = U_{v,alpha*beta} and the proof is via an explicit computation of a single coset decomposition of the double coset U(alpha 0;0 1)U as a disjoint union of (alpha t;0 1)U as t runs through coset reps for O_v/alpha.

bwangpj:
claim

PR #659, opened 2025-07-03
GL_2 over finite adeles is restricted product of local GL_2(K_v)s

bwangpj:
This is [Theorem 12.20](https://imperialcollegelondon.github.io/FLT/blueprint/HeckeOperator_project.html#nolean-GL2-finite-adeles) on the decomposition of global GL_2 as a restricted product of local GL_2. It was needed (not strictly necessary, but would make sense to have first) in #584. It also includes definition and statements of [Lemma 12.16](https://imperialcollegelondon.github.io/FLT/blueprint/HeckeOperator_project.html#nolean-compactopen-matrix) on M_2(O_v) as a (compact open) subring of M_2(K_v).

### DIFF
--- a/FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean
+++ b/FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean
@@ -290,15 +290,9 @@ def ContinuousMulEquiv.restrictedProductMatrix {ι : Type*}
     rw [mul_apply, Matrix.mul_apply]
     have h {x : Matrix n n Πʳ (i : ι), [A i, ↑(C i)]} {i : ι} {j k : n} :
         (restrictedProductMatrix.toFun x) i j k = (x j k) i := by
-      simp only [restrictedProductMatrix, Homeomorph.restrictedProductMatrix,
-        Homeomorph.restrictedProductPi, Homeomorph.piCongrRight, Equiv.restrictedProductPi,
-        Equiv.piCongrRight, AddSubmonoid.coe_set_mk, Subsemiring.coe_carrier_toSubmonoid,
-        Subring.coe_toSubsemiring, AddSubsemigroup.coe_set_mk, Set.mem_setOf_eq, Equiv.coe_fn_mk,
-        Equiv.coe_fn_symm_mk, Equiv.toFun_as_coe, Homeomorph.coe_toEquiv,
-        Homeomorph.symm_trans_apply, Homeomorph.homeomorph_mk_coe_symm, Pi.map_apply, mk_apply]
-      rfl
-    simp only [h]
-    rw [Matrix.mul_apply]
+      simp [restrictedProductMatrix, Homeomorph.restrictedProductMatrix,
+        Homeomorph.restrictedProductPi, Equiv.restrictedProductPi, Matrix]
+    simp only [h, Matrix.mul_apply]
     conv_rhs => arg 2; intro _; rw [← mul_apply]
     apply map_sum (RestrictedProduct.evalAddMonoidHom _ _) _ _
       }


bwangpj:
awaiting-review

FLT/QuaternionAlgebra/NumberField.lean
```diff
@@ -51,11 +55,51 @@ variable {F}
 
 namespace IsDedekindDomain
 
+/-- `M_2(O_v)` as a subring of `M_2(F_v)`. -/
+noncomputable def M2.localFullLevel (v : HeightOneSpectrum (𝓞 F)) :
```
kbuzzard, line 59: I don't think we should be making these subrings in this bespoke situation. See [#Is there code for X? > Set.matrix @ 💬](https://leanprover.zulipchat.com/#narrow/channel/217875-Is-there-code-for-X.3F/topic/Set.2Ematrix/near/527826651) for my suggestion of a general definition (and we'll see what the community thinks).

FLT/QuaternionAlgebra/NumberField.lean
```diff
@@ -51,11 +55,51 @@ variable {F}
...
+      }
+
 noncomputable def GL2.localFullLevel (v : HeightOneSpectrum (𝓞 F)) :
     Subgroup (GL (Fin 2) (v.adicCompletion F)) :=
   MonoidHom.range (Units.map
     (RingHom.mapMatrix (v.adicCompletionIntegers F).subtype).toMonoidHom)
 
+theorem M2.localFullLevel.isOpen (v : HeightOneSpectrum (𝓞 F)) :
+    IsOpen (M2.localFullLevel v).carrier :=
+  sorry
```
kbuzzard: I wrote a proof of this (in the correct generality) in the post I linked to above.

FLT/QuaternionAlgebra/NumberField.lean
```diff
@@ -179,6 +223,60 @@ noncomputable def GL2.toAdicCompletion
     GL (Fin 2) (v.adicCompletion F) :=
   Units.map (RingHom.mapMatrix (FiniteAdeleRing.toAdicCompletion v)).toMonoidHom
 
+/-- `GL_2(𝔸_F^∞)` is isomorphic and homeomorphic to the
+restricted product of the local components `GL_2(F_v)`. -/
+noncomputable def GL2.restrictedProduct :
+  GL (Fin 2) (FiniteAdeleRing (𝓞 F) F) ≃ₜ*
+  Πʳ (v : HeightOneSpectrum (𝓞 F)),
+    [(GL (Fin 2) (v.adicCompletion F)), (M2.localFullLevel v).units] :=
```
kbuzzard: Again I feel like we should be proving this in the correct generality. I have slightly lost track of what we have so far, but IIRC we already have things like "matrix of restricted product is restricted product of matrices" and "units of restricted product is restricted product of units" and we should continue in this vein, proving results in the correct generality.

kbuzzard (commented):
Your proofs are going to be useful but I think that all of this should be framed more generally with no reference at all to completions of number fields. The philosophy when formalizing is to write everything in as much generality as you can, so it can be reused later.

FLT/QuaternionAlgebra/NumberField.lean
bwangpj: Got it! The proof I wrote below uses exactly these two facts you mentioned (and which you also linked to in your other comment). I can rewrite it with no reference to number fields and adele rings and perhaps place it in `FLT.Mathlib.Topology.Algebra.RestrictedProduct`.

FLT/QuaternionAlgebra/NumberField.lean
bwangpj: I see, I'll keep track of this and maybe add this to `FLT.Mathlib`. Without this definition of subring it'll be hard to state the result we need cleanly, especially seeing as `Homeomorph.restrictedProductMatrix` basically uses the same definition of 'matrices with entries in a (sub)set'.

FLT/QuaternionAlgebra/NumberField.lean
bwangpj: This will also need the above-mentioned definition of 'matrices with entries in a subring' as a submonoid of the larger matrix ring.

FLT/QuaternionAlgebra/NumberField.lean
kbuzzard: yeah `Homeomorph.restrictedProductMatrix` should probably be refactored to use the construction I'm proposing.

FLT/QuaternionAlgebra/NumberField.lean
kbuzzard: Yeah I agree. If you're happy to take this on then I'll be happy to merge stuff quickly. I think we first need some basic stuff like I suggested in the mathlib thread, and then slicker (i.e. more general) versions of what you're doing here. I'd feel a lot more comfortable working in that generality because it feels like the correct thing to do formally.

bwangpj: Yeah sure! I'll work on it soon :)

bwangpj:
This now depends on mathlib PR [#27190](https://github.com/leanprover-community/mathlib4/pull/27190) on `Set.matrix`, though I've already made the changes here so everything should work once the other PR is merged and mathlib is updated.

Also, one natural thing to do now is to redefine `GL2.localFullLevel v` as `(M2.localFullLevel v).units`, however this will affect quite a few other things. I'm willing to give it a shot though if we decide that it is worth it.

FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean
```diff
@@ -270,6 +270,52 @@ def ContinuousMulEquiv.restrictedProductUnits {ι : Type*}
...
+      Πʳ i, [Matrix n n (A i), ((C i).matrix : Subring (Matrix n n (A i)))] :=
+    let restrictedProductMatrix :
+        Matrix n n (Πʳ i, [A i, C i]) ≃ₜ
+          Πʳ i, [Matrix n n (A i), ((C i).matrix : Subring (Matrix n n (A i)))] :=
+      Homeomorph.symm (Homeomorph.restrictedProductMatrix hCopen)
+  {
+  __ := restrictedProductMatrix
+  map_mul' x y := by
+    ext i j k
+    rw[mul_apply, Matrix.mul_apply]
```
kbuzzard:
```suggestion
    rw [mul_apply, Matrix.mul_apply]
```
and lines 301, 302

FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean
```diff
@@ -270,6 +270,52 @@ def ContinuousMulEquiv.restrictedProductUnits {ι : Type*}
...
+    have h {x : Matrix n n Πʳ (i : ι), [A i, ↑(C i)]} {i : ι} {j k : n} :
+        (restrictedProductMatrix.toFun x) i j k = (x j k) i := by
+      simp only [restrictedProductMatrix, Homeomorph.restrictedProductMatrix,
+        Homeomorph.restrictedProductPi, Homeomorph.piCongrRight, Equiv.restrictedProductPi,
+        Equiv.piCongrRight, AddSubmonoid.coe_set_mk, Subsemiring.coe_carrier_toSubmonoid,
+        Subring.coe_toSubsemiring, AddSubsemigroup.coe_set_mk, Set.mem_setOf_eq, Equiv.coe_fn_mk,
+        Equiv.coe_fn_symm_mk, Equiv.toFun_as_coe, Homeomorph.coe_toEquiv,
+        Homeomorph.symm_trans_apply, Homeomorph.homeomorph_mk_coe_symm, Pi.map_apply, mk_apply]
+      exact rfl
```
kbuzzard:
```suggestion
      rfl
```

kbuzzard (commented):
Modulo the nits I point out, this looks fine. Thanks! Can you reply `awaiting-review` when you've dealt with them, and I'll merge.

FLT/Mathlib/Topology/Algebra/RestrictedProduct/TopologicalSpace.lean
```diff
@@ -270,6 +270,52 @@ def ContinuousMulEquiv.restrictedProductUnits {ι : Type*}
...
+    rw [mul_apply, Matrix.mul_apply]
+    have h {x : Matrix n n Πʳ (i : ι), [A i, ↑(C i)]} {i : ι} {j k : n} :
+        (restrictedProductMatrix.toFun x) i j k = (x j k) i := by
+      simp only [restrictedProductMatrix, Homeomorph.restrictedProductMatrix,
+        Homeomorph.restrictedProductPi, Homeomorph.piCongrRight, Equiv.restrictedProductPi,
+        Equiv.piCongrRight, AddSubmonoid.coe_set_mk, Subsemiring.coe_carrier_toSubmonoid,
+        Subring.coe_toSubsemiring, AddSubsemigroup.coe_set_mk, Set.mem_setOf_eq, Equiv.coe_fn_mk,
+        Equiv.coe_fn_symm_mk, Equiv.toFun_as_coe, Homeomorph.coe_toEquiv,
+        Homeomorph.symm_trans_apply, Homeomorph.homeomorph_mk_coe_symm, Pi.map_apply, mk_apply]
+      rfl
```
kbuzzard: `simp` then `rfl` always looks so inelegant, because you've been forced to squeeze the `simp`. You're right that `simp` doesn't close the goal but `rfl` does, so how to fix this? Here's my thought process.  My first instinct is "we're missing a `simp` lemma". So look at the goal and it's `(RestrictedProduct.mk blah) (input)` so there should be `RestrictedProduct.mk_apply`. But this lemma already exists and it's tagged `simp`. So I tried `rw` and the rewrite fails, so this is the problem. I then tried `erw [RestrictedProduct.mk_apply]` and this works! So I tried `erw? [RestrictedProduct.mk_apply]` to see why `rw` fails, and the output in the infoview seems to indicate that `Matrix` isn't unfolding. So then I tried `simp [Matrix]` and this closes the goal. So can you delete `rfl` and add `Matrix` to the list of things in the `simp only` call, and then change it to `simp` and then delete all the stuff you don't need now, and your proof now looks much cleaner:

```lean
    have h {x : Matrix n n Πʳ (i : ι), [A i, ↑(C i)]} {i : ι} {j k : n} :
        (restrictedProductMatrix.toFun x) i j k = (x j k) i := by
      simp [restrictedProductMatrix, Homeomorph.restrictedProductMatrix, Matrix,
        Homeomorph.restrictedProductPi, Equiv.restrictedProductPi]
    simp only [h, Matrix.mul_apply]
```

bwangpj: Ah thank you very much! Yeah I struggled for a while to make this part of the proof shorter but it didn't occur to me to try `erw?` - this is indeed useful!

kbuzzard:
Thanks!

PR #666, opened 2025-07-08
feat: add CITATION.bib

kbuzzard:


pitmonticone (approved):
LGTM! 

I think that in our case we should have `year = {2025}` and `version = {v4.22.0-rc3}` and then I’ll automate the update when I find the time. It should be quite quick.

But for the moment the one you added here is more than enough. I think it makes send to specify the version iff we have automation that deals with version bumping.

PR #674, opened 2025-07-13
chore: replace Subtype.map_ne

Ruben-VandeVelde:
As suggested by Jireh in https://github.com/leanprover-community/mathlib4/pull/27012


kbuzzard:
Thanks!

PR #693, opened 2025-08-11
feat: WeierstrassCurve.n_torsion_dimension

hanwenzhu:
This PR proves a simple theorem in EllipticCurve/Torsion.lean

(The proof was found by Seed-Prover at its lightweight setting and then manually minimized by me.)

### DIFF
--- a/FLT/EllipticCurve/Torsion.lean
+++ b/FLT/EllipticCurve/Torsion.lean
@@ -52,8 +52,85 @@ theorem group_theory_lemma {A : Type*} [AddCommGroup A] {n : ℕ} (hn : 0 < n) (
 -- I only need this if n is prime but there's no harm thinking about it in general I guess.
 -- It follows from the previous theorem using pure group theory (possibly including the
 -- structure theorem for finite abelian groups)
+
+/- Start of proof attempt -/
+lemma round1_h1 (k : Type u) [Field k] (E : WeierstrassCurve k) (n : ℕ) (hn : (n : k) ≠ 0) : 0 < n := by
+  by_contra h2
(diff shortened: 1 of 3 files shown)


FLT/EllipticCurve/Torsion.lean
```diff
@@ -53,7 +53,15 @@ theorem group_theory_lemma {A : Type*} [AddCommGroup A] {n : ℕ} (hn : 0 < n) (
 -- It follows from the previous theorem using pure group theory (possibly including the
 -- structure theorem for finite abelian groups)
 theorem WeierstrassCurve.n_torsion_dimension [IsSepClosed k] {n : ℕ} (hn : (n : k) ≠ 0) :
-    ∃ φ : E.n_torsion n ≃+ (ZMod n) × (ZMod n), True := sorry
+    ∃ φ : E.n_torsion n ≃+ (ZMod n) × (ZMod n), True := by
```
alreadydone: The statement should probably be changed to `Nonempty (E.n_torsion n ≃+ (ZMod n) × (ZMod n))` (same for the previous one)

hanwenzhu: @alreadydone or maybe change it to a `def`? I'm a little hesitant to change the statement myself because they weren't written by me, and I don't completely understand the context.

alreadydone: I don't think there's a canonical isomorphism so wrapping it in Nonempty seems more appropriate :)

hanwenzhu: The code is changed as suggested. Thanks!

Issue #243, opened 2024-11-26
Proof that L (x)[K] A_K^f = A_L^f


kbuzzard:
Base change of finite adeles of K is topologically and algebraically isomorphic to finite adeles of L. Can't even state this yet because we need continuous algebra equivs (see

maddycrim:
claim

kbuzzard:
@maddycrim I put some more details on my proposed proof in the LaTeX file in my branch `kbuzzard-adele-refactor`, currently a WIP PR #399 while we wait for the finite adele refactor to hit mathlib. If you know a better proof then I'm happy to change plan here. The reason I'm pinging you is that Javier Lopez had been working on this but now he got a new job and has no time to work on it any more :-/ His work is pushed in WIP PR #416 and there are still plenty of sorrys.

kbuzzard:
I now think that the way to do this is to prove 
(a) tensoring with a finite module commutes with restricted products
(b) some general nonsense about restricted products.

I think (a) follows from
(a1) tensoring with a finite module commutes with arbitrary products;
(a2) tensoring with a finite module commutes with filtered colimits.

These statements need to be formalized.

PR #697, opened 2025-08-14
Mostly finish showing that adele base change map is an isomorphism

matthewjasper:
Reduces bijectivity of FiniteAdeleRing.baseChangeAlgEquiv to a result about tensor products commuting with restricted products.

Most of #243


kbuzzard:
Thanks so much for this. I feel like you know this part of the FLT library better than me!

PR #660, opened 2025-07-04
Explicit local double coset decomposition in preparation for Hecke operators

bwangpj:
This is [Lemma 12.19](https://imperialcollegelondon.github.io/FLT/blueprint/HeckeOperator_project.html#nolean-U1-coset-decomposition) of the blueprint, which is the essential component of #584. It is in a separate file due to its length, but also because it is a purely local computation in preparation for the commutativity of Hecke operators, and there didn't seem to be anywhere else suitable to place it.

PR #711, opened 2025-09-05
feat: commutativity of Hecke operators at bad primes modulo two sorries

bwangpj:
This PR accomplishes two things:

- refactor terms of the form `(U * g • V) : Set (G ⧸ V)` to `(U * {g}) : Set (G ⧸ V)`;
- reduce the proof of the commutativity of Hecke operators at bad primes `U_mul` to two groups of (easier) sorries:
    - `bijOn_unipotent_mul_diagU1_U1diagU1`: global double coset decomposition, which should be able to be deduced from the local double coset decomposition of #660 
    - `U_mul_aux`: should be a fairly straightforward lemma, relating the sum over O_v/alpha, O_v/beta to the sum over O_v/(alpha*beta).

### DIFF
--- a/FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Abstract.lean
+++ b/FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Abstract.lean
@@ -123,7 +123,7 @@ open MulAction
 
 -- finiteness hypothesis we need to make Hecke operators work: UgV is
 -- a finite number of left V-cosets.
-variable (h : (QuotientGroup.mk '' (U * g • V) : Set (G ⧸ V)).Finite)
+variable (h : (QuotientGroup.mk '' (U * {g}) : Set (G ⧸ V)).Finite)
 
 open ConjAct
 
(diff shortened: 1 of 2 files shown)


FLT/AutomorphicForm/QuaternionAlgebra/HeckeOperators/Abstract.lean
```diff
@@ -182,21 +184,24 @@ noncomputable def HeckeOperator : fixedPoints V A →ₗ[R] fixedPoints U A wher
...
     ext
-    simp [HeckeOperator_toFun, -Set.mem_image, finsum_mem_add_distrib (h.image Quotient.out)]
+    simp only [HeckeOperator_toFun, FixedPoints.coe_add, smul_add,
+      finsum_mem_add_distrib (h.image Quotient.out)]
   map_smul' r a := by
     ext
-    simp [-Set.mem_image, HeckeOperator_toFun, smul_comm, smul_finsum_mem (h.image Quotient.out)]
+    simp only [HeckeOperator_toFun, FixedPoints.coe_smul, smul_comm,
+      smul_finsum_mem (h.image Quotient.out)]
+    rfl
```
kbuzzard: `simp` was closing this goal before -- does it not close it now? If not then maybe `simp` is not confluent and there's a missing lemma? If so then squeeze it and add whatever is finishing the job to line 191.

bwangpj:
awaiting-review

PR #723, opened 2025-09-17
chore: shake imports

Ruben-VandeVelde:
I'm not sure if any of these are intended to stay in the expectation that they'll be used to prove `sorry`s.

I moved the `IsLocalHom e.toRingHom` instance to prevent Lean from generating the same name for two of these similar instances.


kbuzzard:
Thanks!

PR #727, opened 2025-09-19
Slow import issue

kbuzzard:
Debugging some slowness with imports. Not for merging. File FLT/GaloisRepresentation/crap/GaloisGroup_for_Oliver.lean currently doesn't compile (but is also not mentioned in `FLT.lean` so `lake build` works). There is probably some dodgy instance in FLT which we want to find. You can fix the file by commenting out line 1 `import FLT.Deformations.RepresentationTheory.Basic`, so do this, check that it compiles, and then uncomment line 1 again, check the file now doesn't compile, and then jump to `FLT.Deformations.RepresentationTheory.Basic` and see what the FLT-imports to that file are, and then replace the import by these new imports.

My problem is that (at least if I am not connected to the internet, I don't know if this makes it worse) then the procedure I outlined above takes over a minute because every time I change the imports I find myself in orange bar hell.

### DIFF
--- a/FLT/GaloisRepresentation/crap/GaloisGroup_for_Oliver.lean
+++ b/FLT/GaloisRepresentation/crap/GaloisGroup_for_Oliver.lean
@@ -1,0 +1,42 @@
+import FLT.Deformations.RepresentationTheory.Basic
+import Mathlib
+
+-- The problem: uncommenting the FLT import on line 1 breaks the proof.
+-- Show this by uncommenting the FLT import and then commenting it out again, and then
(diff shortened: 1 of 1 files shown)


kbuzzard:
Resolved -- this was caused by Microsoft Defender. Zulip thread [#general > building mathlib on ubuntu v mac @ 💬](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/building.20mathlib.20on.20ubuntu.20v.20mac/near/553257509)

PR #735, opened 2025-10-03
feat: statement of cyclic base change

kbuzzard:
Statement of cyclic base change which covers all the cases we need.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -79,7 +79,7 @@ jobs:
       - name: Upstreaming dashboard
         run: python3 scripts/upstreaming_dashboard.py
 
-      - uses: leanprover-community/docgen-action@095763bcfa35bef9c6a3eb8ae778c5e6c7727df2 # 2025-07-03
+      - uses: leanprover-community/docgen-action@deed0cdc44dd8e5de07a300773eb751d33e32fc8 # 2025-10-26
         with:
           blueprint: true
           homepage: docs


FLT/GaloisRepresentation/Automorphic.lean
```diff
@@ -57,35 +60,116 @@ This is a far more restrictive definition of automorphic than is found in the li
...
-    (D : Type*) [Ring D] [Algebra F D] [IsQuaternionAlgebra F D]
-    -- `D` is assumed to have discriminant 1
-    (r : IsQuaternionAlgebra.NumberField.Rigidification F D)
     -- `S` is the level of the modular form
     (S : Finset (HeightOneSpectrum (𝓞 F))) : Prop :=
-  -- We say `ρ` is *automorphic* if there's an `A`-valued automorphic eigenform
-  ∃ (π : HeckeAlgebra F D r S 𝒪 →ₐ[𝒪] A),
+  -- We say `ρ` is *automorphic* if there's a quaternion algebra D over F of discriminant 1
+  ∃ (D : Type u) (_ : Ring D) (_ : Algebra F D) (_ : IsQuaternionAlgebra F D)
+    (r : IsQuaternionAlgebra.NumberField.Rigidification F D)
```
erdOne, line 78: Do we really want to move this into the exists? This gives you a different `D` and rigidification whenever you obtain from it and you need either a theorem or a sorry showing that such `D` is unique to actually use it. 

But the sorry is definitely known in the 80s so it's probably fine either way.

FLT/GaloisRepresentation/Automorphic.lean
```diff
@@ -57,35 +60,116 @@ This is a far more restrictive definition of automorphic than is found in the li
...
+
+instance {F E D : Type*}
+    [Field F]
+    [Field E] [Algebra F E]
+    [Ring D] [Algebra F D] [IsQuaternionAlgebra F D] :
+    IsQuaternionAlgebra E (E ⊗[F] D) := sorry -- Ask Edison?
+
+variable (p : ℕ) [Fact p.Prime] in
+instance : ContinuousSMul ℤ_[p] (AlgebraicClosure ℚ_[p]) where
+  continuous_smul := sorry -- ask on is there code for X
```
erdOne: This is 
```
variable {p : ℕ} [Fact p.Prime]

noncomputable
instance : NormedSpace ℚ_[p] (PadicAlgCl p) := spectralNorm.normedSpace ..

instance : ContinuousSMul ℤ_[p] (PadicAlgCl p) :=
  continuousSMul_of_algebraMap _ _ ((continuous_algebraMap ℚ_[p] _).comp continuous_subtype_val)
```

FLT/GaloisRepresentation/Automorphic.lean
```diff
@@ -57,35 +60,116 @@ This is a far more restrictive definition of automorphic than is found in the li
...
+* `det(ρ)` is the cyclotomic character.
+
+Then `ρ` is automorphic of level U₁(S) iff ρ|G_E is automorphic of level U₁(S_E),
+where S_E is the pullback of S to E.
+-/
+theorem cyclic_base_change
+    -- let F be a totally real number field of even degree
+    {F : Type*} [Field F] [NumberField F] [IsTotallyReal F]
+    (hF : Even (Module.finrank ℚ F))
+    -- let E/F be a finite solvable extension
```
erdOne, line 126: You also need either `Even (Module.finrank ℚ E)` or `Even (Module.finrank F E)` I think?

FLT/GaloisRepresentation/Automorphic.lean
kbuzzard: The reason I moved D inside was so I could state cyclic base change as "rho is automorphic of level S iff rho|E is automorphic of level S|E" rather than having to introduce two quaternion algebras in the statement (and then you'll get different quaternion algebras whenever you use cyclic base change, so it's basically the same problem). Another possibility is "for all D/F and for all rigidifications r,..." (after the colon), and then you run into the problem that you will sometimes need to produce a D and again this is known in the 80s. I propose just leaving it as it is and adding lemmas as we need them. Does this sound OK?

FLT/GaloisRepresentation/Automorphic.lean
kbuzzard: Thanks!

FLT/GaloisRepresentation/Automorphic.lean
kbuzzard: I don't think we need F/E even, and E/Q even follows from F/Q even. Does this sound OK?

FLT/GaloisRepresentation/Automorphic.lean
erdOne: Ah sure makes sense.

PR #738, opened 2025-10-03
feat (Fujisaki): ringHaarChar of modules iso to \R^d.

WilliamCoram:
We show that for a module R over the reals, the ringHaarChar of a unit of the reals in R, can be calculated by considering R as isomorphic to reals^dim.

### DIFF
--- a/FLT/DivisionAlgebra/HaarCharWork.lean
+++ b/FLT/DivisionAlgebra/HaarCharWork.lean
@@ -155,7 +155,7 @@ abbrev iso : R ≃ₜ+ (Fin (Module.finrank ℝ R) → ℝ) where
 
 variable [IsTopologicalRing R] [LocallyCompactSpace R] [MeasurableSpace R] [BorelSpace R]
 
-lemma ringHaarChar_eq1' (y : (Fin (Module.finrank ℝ R) → ℝ)ˣ)
+lemma ringHaarChar_eq (y : (Fin (Module.finrank ℝ R) → ℝ)ˣ)
     (hy : ∃ a : ℝ, y.val = algebraMap ℝ (Fin (Module.finrank ℝ R) → ℝ) a)
     (hy' : IsUnit ((Module.Basis.equivFun
       (Classical.choose (ex_basis_sum_basis_eq_one R))).symm y)) :
--- a/FLT/DivisionAlgebra/HaarCharWork.lean
+++ b/FLT/DivisionAlgebra/HaarCharWork.lean
@@ -1,0 +1,179 @@
+import FLT.HaarMeasure.HaarChar.Ring
+import Mathlib.Algebra.Lie.OfAssociative
+import Mathlib.Analysis.Normed.Lp.lpSpace
+import Mathlib.Order.CompletePartialOrder
+
+variable (R : Type*) [Ring R] [Nontrivial R] [Algebra ℝ R] [Module.Finite ℝ R] [Module.Free ℝ R]
+  [TopologicalSpace R] [IsModuleTopology ℝ R]
+
+namespace ringHaarChar.Module
+
+/-- For a basis, there exists a non-zero coefficient of a basis element in the expansion of one. -/
+def ex_ne_zero_coeff_one (B : Module.Basis (Fin (Module.finrank ℝ R)) ℝ R) :
+  ∃ i, B.coord i 1 ≠ 0 := by
+  by_contra t
+  simp only [Module.Basis.coord_apply, ne_eq, not_exists, Decidable.not_not] at t
+  have w (a b : R) : (∀ x, (B.coord x) a = (B.coord x) b) →
+      (a = b) := (Module.Basis.ext_elem_iff B).mpr
+  specialize w 1 0
+  have (x : Fin (Module.finrank ℝ R)) : (B.coord x) 1 =
+      (B.coord x) 0 := by
+    simpa using t x
+  apply w at this
+  contrapose this
+  exact one_ne_zero
+
+/-- The basis change map mapping `B i ↦ B i - (∑ x, B x - 1)` where `i` is from
+  `ex_ne_zero_coeff_one`. -/
+noncomputable
+def basis_change_map :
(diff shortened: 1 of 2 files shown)


FLT/DivisionAlgebra/HaarCharWork.lean
```diff
@@ -0,0 +1,179 @@
+import FLT.HaarMeasure.HaarChar.Ring
```
kbuzzard: This should not be called `HaarCharWork` and this should not be in `DivisionAlgebra`. Probably it should be somewhere in `FLT/Mathllib/`.

FLT/DivisionAlgebra/HaarCharWork.lean
```diff
@@ -0,0 +1,179 @@
...
+import Mathlib.Algebra.Lie.OfAssociative
+import Mathlib.Analysis.Normed.Lp.lpSpace
+import Mathlib.Order.CompletePartialOrder
+
+variable (R : Type*) [Ring R] [Nontrivial R] [Algebra ℝ R] [Module.Finite ℝ R] [Module.Free ℝ R]
+  [TopologicalSpace R] [IsModuleTopology ℝ R]
```
kbuzzard: Too early.

FLT/DivisionAlgebra/HaarCharWork.lean
```diff
@@ -0,0 +1,179 @@
...
+
+namespace ringHaarChar.Module
+
+omit [Module.Finite ℝ R] [Module.Free ℝ R] [TopologicalSpace R] [IsModuleTopology ℝ R] in
```
kbuzzard: See? ;-)

FLT/DivisionAlgebra/HaarCharWork.lean
WilliamCoram: The name and location was never meant to be correct (as with the other file)- I was trying to keep it all together for myself until you are happy with the actual content, then I can worry with where it belongs and what it should be called. 
I can try and sort this out first in the future (and now for these files) if that is something you would prefer though

FLT/DivisionAlgebra/HaarCharWork.lean
WilliamCoram: I'll sort the variables in this and the other files out before doing anything else. I see how this is bad practice now.

FLT/DivisionAlgebra/HaarCharWork.lean
kbuzzard: I see.

If you make a PR and mark it `awaiting review` then to me this means "this is ready to go in the FLT repo" and in particular it means "I believe that everything is in the right place, everything is named correctly etc etc, this is ready to become an official part of the proof of Fermat's Last Theorem". I can't be accruing technical debt by merging things which aren't completely ready. I now understand that you are treating the PR process differently -- you just want me to look at your code and you don't even care that the files are named completely incorrectly etc. What you're trying to do is ask me whether your code/approach is OK. I had not really appreciated this.

kbuzzard:
I think that what you should be proving is the following.

```
variable (R : Type*) [Ring R] [Algebra ℝ R] [Module.Finite ℝ R]
    [TopologicalSpace R] [IsTopologicalRing R] [IsModuleTopology ℝ R]
    [LocallyCompactSpace R] [MeasurableSpace R] [BorelSpace R]
    (t : ℝˣ)

theorem foo : MeasureTheory.ringHaarChar (Units.map (algebraMap ℝ R).toMonoidHom t) =
    ‖(t : ℝ)‖₊ ^ (Module.finrank ℝ R) := sorry
```

You should prove this in the following steps.

1) Show that the LHS is equal to `addEquivAddHaarChar` of `v ↦ t • v` considered as a continuous automorphism of R as an additive group (you should find the construction in mathlib giving the continuous automorphism of a topological module induced by multiplying by a unit of the topological ring that it's a module over; ask on Zulip if you can't find it, but don't make yourself until you're convinced it's not there). This should just be a matter of unfolding the definitions.

2) You should use `MeasureTheory.addEquivAddHaarChar_eq_addEquivAddHaarChar_of_continuousAddEquiv` and the continuous isomorphism `R ≃ₜ+ (Fin (Module.finrank ℝ R) → ℝ`) (which will surely already be in mathlib; ask on the Zulip if you can't find it; continuity comes `IsModuleTopology.instPi`) to reduce the question to showing that the addEquivAddHaarChar of multiplication by a nonzero real on `Fin d → ℝ` is `‖(t : ℝ)‖₊ ^ d`

3) You should then deduce the result from `MeasureTheory.addEquivAddHaarChar_piCongrRight` and `MeasureTheory.ringHaarChar_real`.

FLT/DivisionAlgebra/HaarCharWork.lean
WilliamCoram: Yeah, that is how I treat the PR process. In my mind the naming and finding files is the "easy part", although this is not always true, and so comes at the end of the process.

WilliamCoram: Perhaps we need an interim label, something like "needs correspondence" (too formal?) which stands for "maths and lean needs checking"?

WilliamCoram:
I found my confusion. In my set up when I required ``(t,t,...t)`` to be t inside T^d I was inherently forcing a choice of basis.

I was doing this was because I was in the revserse mindset. That is I was taking an element of T to an element of T^d (diagonal embedding) then taking it to R, and similarly working out their ringHaarChar at each step. Where as I should have been working from inside R first, then mapping across without any assumptions.

FLT/HaarMeasure/HaarChar/Ring.lean
```diff
@@ -195,4 +196,44 @@ lemma ringHaarChar_restrictedProduct (u : (Πʳ i, [A i, C i])ˣ) :
 
 end restrictedproduct
 
+section ModuleFinite
+
+variable {T R : Type*} [Field T] [Ring R] [Algebra T R] [Module.Finite T R]
```
kbuzzard: Can you call the field K? Thanks :-)

PR #741, opened 2025-10-09
Taylor Kim note chapter 01

bourbaki120205:
This is part of the formalization of Taylor Kim notes chapter 01.


FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+theorem FLTfour : ∀ a b c : ℕ , a ≠ 0 → b ≠ 0 → c ≠ 0 → a ^ 4 + b ^ 4 ≠ c ^ 4 :=
+fermatLastTheoremFour
+
+/-- we hence only need to care about n=ℓ an odd prime larger than 3,
+ we can assume we have a Frey's package in this case -/
+lemma reduce {a b c : ℤ} (ha : a ≠ 0) (hb : b ≠ 0) (hc : c ≠ 0)
+    {p : ℕ} (pp : p.Prime) (hp5 : 5 ≤ p) (H : a ^ p + b ^ p = c ^ p) : Nonempty FLT.FreyPackage :=
+    FLT.FreyPackage.of_not_FermatLastTheorem_p_ge_5 ha hb hc pp hp5 H
+
+def IsBadReduction (C : WeierstrassCurve ℚ) (p : ℕ) [Fact (Nat.Prime p)] :Prop:= sorry
```
kbuzzard, line 27: It would be an interesting project to write down this definition.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+
+variable (p : ℕ) [Fact (Nat.Prime p)]
+
+lemma badreduction (P : FreyPackage) :
+IsBadReduction (FreyCurve P) p ↔ ((p : Int) ∣ (P.a * P.b * P.c)):= sorry
```
kbuzzard, line 32: It would also be an interesting project to prove this theorem! Both ways are probably possible -- you could first prove that if the valuation of the j-invariant is negative then the curve has bad reduction, and then for p coprime to abc you could explicitly check that it has good reduction. This would teach you something about eliptic curves and probably about p-adic fields as well.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+
+local notation "G_K" => absoluteGaloisGroup
```
kbuzzard, line 34: I think `\Gamma` is better notation, I don't think notation should have random letters like K in it.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+
+/-noncomputable instance Modulestructue (C : WeierstrassCurve ℚ) :
+Module (ℤ[G] (G_K ℚ_[p])) (E_pclosure C p) := sorry-/
+
+abbrev G (q : IsLocalRing.maximalIdeal ℤ_[p]) (hq : (q : ℚ_[p]) ≠ 0) :
+ Type := Additive (ℚ_[p]ˣ⧸(Subgroup.closure ({Units.mk0 (q: ℚ_[p]) hq})))
+
+noncomputable instance AddCommMonoid
+(q : IsLocalRing.maximalIdeal ℤ_[p]) (hq : (q : ℚ_[p]) ≠ 0) : AddCommMonoid (G p q hq)
+:= by infer_instance
```
kbuzzard, line 53: If the proof is `by infer_instance` then you don't need it in the first place.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+structure Representation.Equiv {R : Type*} [CommSemiring R] {G : Type*} [Group G]
+    {M₁ M₂ : Type*} [_root_.AddCommMonoid M₁] [_root_.AddCommMonoid M₂] [Module R M₁] [Module R M₂]
+    (ρ₁ : Representation R G M₁) (ρ₂ : Representation R G M₂)
+     extends M₁ ≃ₗ[R] M₂ where
+     exchange : ∀(g : G), ∀(m : M₁), toFun (ρ₁ g m) = (ρ₂ g (toFun m))
+
+def twist (ρ : Representation ℤ G M) (δ : G →* ℤˣ) : Representation ℤ G M where
+  toFun g := {
+    toFun m := δ g • ρ g m
+    map_add' := by aesop
```
kbuzzard, line 73: This is probably quite slow. Does `grind` do it? Can you just prove it easily by hand?

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+    map_smul' := by {
+    intro m x
+    simp only [map_smul, eq_intCast, Int.cast_eq]
+    exact smul_comm (δ g) m ((ρ g) x)
+    }}
+  map_one' := by aesop
```
kbuzzard, line 79: Same comment here

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+  map_mul' g h := by {
+    ext m
+    simp only [map_mul, Module.End.mul_apply, LinearMap.coe_mk, AddHom.coe_mk, map_zsmul_unit]
+    rw[mul_comm, mul_smul]
```
kbuzzard, line 83: Can these go in the simp call?

kbuzzard, line 83: Can these go in the simp call?

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+/-This theorem can be even strengthen as topological module isomorphism, I'll write later-/
+/-what's the difference between a Z tensor and a Z[G] tensor?-/
+theorem Tate (P : FreyPackage) : IsBadReduction (FreyCurve P) p
+→ ∃ (δ : (G_K ℚ_[p]) →* ℤˣ) , ∃ (q : IsLocalRing.maximalIdeal ℤ_[p])
+(hq : (q : ℚ_[p]) ≠ 0), Nonempty (Representation.Equiv (ρ_curve p (FreyCurve P))
+(twist (ρ_G p q hq) δ)) := sorry
```
kbuzzard, line 91: You might want to define both split and nonsplit multiplicative reduction (which tells you what delta is) and state the correct theorem (split reduction => iso exists, not delta needed, and nonsplit => needs delta)

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+
+variable (q : IsLocalRing.maximalIdeal ℤ_[p]) (hq : (q : ℚ_[p]) ≠ 0) {P : FreyPackage}
+(hPp : IsBadReduction (FreyCurve P) p) (δ : (G_K ℚ_[p]) →* ℤˣ)
+(hqδ : Nonempty (Representation.Equiv (ρ_curve p (FreyCurve P))
+(twist (ρ_G p q hq) δ)))
+
+section
+include q hq P hPp δ hqδ
+lemma q_equation : ∃(C : ℕ → ℚ_[p]), (FreyCurve P).j=
```
kbuzzard, line 100: C takes values in Z! And you might just want to remove the 744 and sum over all of Nat, and then just put in a lemma saying C 0 = 744.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+(q : ℚ_[p])⁻¹ + 744 + ∑' n : ℕ+, (C n) * q^(n:ℕ) := sorry
+
+lemma p_valuation : Padic.valuation ((FreyCurve P).j : ℚ_[p])
+ = Padic.valuation (q : ℚ_[p])⁻¹ := sorry
```
kbuzzard, line 104: This says "for all q, v(j)=v(q)^{-1} as far as I can see.

FLT/TaylorKimNotes/Chapter01.lean
```diff
@@ -0,0 +1,129 @@
...
+
+lemma p_valuation_mod (hp : Odd p) :
+((Padic.valuation (p := p ) (q:ℤ_[p])) : ZMod (2*P.p)) = 0 := sorry
+
+lemma p_valuation_mod2 : ((Padic.valuation (p := p ) (q:ℤ_[p])) : ZMod 2) = -8 := sorry
```
kbuzzard, line 109: Similarly this says "for all q, the p-adic valuation of q, when reduced mod 2, is -8", which is surely far from what you want to say.

FLT/TaylorKimNotes/test2.lean
```diff
@@ -0,0 +1,25 @@
+import Mathlib
+
+variable {G M : Type*} [Group G] [AddCommGroup M]
+
+structure Representation.Equiv {R : Type*} [CommSemiring R] {G : Type*} [Group G]
+    {M₁ M₂ : Type*} [AddCommMonoid M₁] [AddCommMonoid M₂] [Module R M₁] [Module R M₂]
+    (ρ₁ : Representation R G M₁) (ρ₂ : Representation R G M₂)
+     extends M₁ ≃ₗ[R] M₂ where
+     exchange : ∀(g : G), ∀(m : M₁), toFun (ρ₁ g m) = (ρ₂ g (toFun m))
```
kbuzzard, line 9: You might want to PR this to mathlib

FLT/TaylorKimNotes/test2.lean
```diff
@@ -0,0 +1,25 @@
...
+
+def twist (ρ : Representation ℤ G M) (δ : G →* ℤˣ) : Representation ℤ G M where
```
kbuzzard, line 11: This too! In the right generality (general base commring R)

FLT/TaylorKimNotes/Chapter01.lean
bourbaki120205-cell: It seems that grind doesn’t work here, I changed to a simpler version by hand

kbuzzard:
(closed due to inactivity)

PR #742, opened 2025-10-10
feat: add theorem statements about hardly ramified Galois representations

kbuzzard:
Add various statements about hardly ramified Galois representations in Lean and glue to LaTeX


FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -77,29 +77,25 @@ instance : MulAction (Γ ℚ_[2]) Z2bar where
   one_smul z := rfl
   mul_smul g h z := rfl
 
-/-- Let `R` be a "local pro-artinian algebra" (for example any complete Noetherian local ring
-with the maximal ideal-adic topology) having finite residue field of characteristic `ℓ > 2`,
-and let `ρ : Gal(Qbar/Q) → GL_2(R)` be a continuous 2-dimensional representation.
-We say that `ρ` is *hardly ramified* if it has cyclotomic determinant, is unramified outside `2ℓ`,
-flat at `ℓ` and upper-triangular at 2 with a 1-dimensional quotient which is unramified and
-whose square is trivial. -/
+/-- Let `R` be a compact Hausdorff local toppologcal ring (for example any complete Noetherian
```
erdOne:
```suggestion
/-- Let `R` be a compact Hausdorff local topologcal ring (for example any complete Noetherian
```

FLT/GaloisRepresentation/HardlyRamified/Family.lean
```diff
@@ -0,0 +1,54 @@
...
+open scoped TensorProduct
+
+universe u v
+
+-- let ρ : G_ℚ → GL_2(R) be hardly ramified, where R is the integers in a finite
+-- extension of ℚ_p
+variable {p : ℕ} (hpodd : Odd p) [hp : Fact p.Prime]
+    {R : Type u} [CommRing R] [Algebra ℤ_[p] R] [IsDomain R]
+    [Module.Finite ℤ_[p] R] [TopologicalSpace R] [IsTopologicalRing R]
+    [IsLocalRing R] [IsModuleTopology ℤ_[p] R]
```
erdOne, line 17: Does the m-adic topology coincide with the Z_p-module topology for any local ring finite over Z_p?

FLT/GaloisRepresentation/HardlyRamified/Family.lean
```diff
@@ -0,0 +1,54 @@
...
+    σ.isCompatible ∧
+    -- are "hardly ramified" for ℓ>2,
+    (∀ {ℓ : ℕ} (hℓ : Fact ℓ.Prime) (hℓodd : Odd ℓ) (φ : E →+* AlgebraicClosure ℚ_[ℓ]),
+      -- by which we mean that for a representation σ_φ in the family,
+      -- there's a hardly-ramified representation `τ` to GL_2(A)
+      -- for A a module-finite free ℤ_ℓ-algebra
+      ∃ (A : Type u) (_ : CommRing A) (_ : TopologicalSpace A) (_ : IsTopologicalRing A)
+        (_ : IsLocalRing A) (_ : Algebra ℤ_[ℓ] A) (_ : Module.Finite ℤ_[ℓ] A)
+        (_ : Module.Free ℤ_[ℓ] A) (_ : IsDomain A) (_ : Algebra A (AlgebraicClosure ℚ_[ℓ]))
+        (_ : ContinuousSMul A (AlgebraicClosure ℚ_[ℓ]))
```
erdOne, line 37: Do you need either `IsScalarTower ℤ_[ℓ] A (AlgebraicClosure ℚ_[ℓ])` or `ContinuousSMul ℤ_[ℓ] A` or `IsAdicTopology A`? Or are they all automatic?

FLT/GaloisRepresentation/HardlyRamified/Family.lean
```diff
@@ -0,0 +1,54 @@
...
+variable {p : ℕ} (hpodd : Odd p) [hp : Fact p.Prime]
+    {R : Type u} [CommRing R] [Algebra ℤ_[p] R] [IsDomain R]
+    [Module.Finite ℤ_[p] R] [TopologicalSpace R] [IsTopologicalRing R]
+    [IsLocalRing R] [IsModuleTopology ℤ_[p] R]
+    {V : Type v} [AddCommGroup V] [Module R V] [Module.Finite R V]
+    [Module.Free R V] (hv : Module.rank R V = 2) {ρ : GaloisRep ℚ R V}
+    (hρ : IsHardlyRamified hpodd hv ρ)
+
+-- Then `ρ` lives in a compatible family of Galois representations
+theorem mem_isCompatible :
```
erdOne: I've not checked myself but you might want to make sure if `hρ` is actually `include`d or not.

FLT/GaloisRepresentation/HardlyRamified/ModThree.lean
```diff
@@ -0,0 +1,20 @@
...
+/-- A mod 3 hardly ramified representation is an extension of trivial by cyclo -/
+-- Probably `Field k` can be replaced with `(3 : k) = 0`
+theorem mod_three {k : Type u} [Fintype k] [Field k] [Algebra ℤ_[3] k] --
+    [TopologicalSpace k] [DiscreteTopology k]
+    (V : Type*) [AddCommGroup V] [Module k V] [Module.Finite k V] [Module.Free k V]
+    (hV : Module.rank k V = 2) {ρ : GaloisRep ℚ k V}
+    (hρ : IsHardlyRamified (show Odd 3 by decide) hV ρ) :
+    ∃ (π : V →ₗ[k] k) (_ : Function.Surjective π),
+    ∀ g : Γ ℚ, ∀ v : V, π (ρ g v) = π v := by
+  sorry
```
erdOne, line 18: Does this say that the kernel is cyclo? Or is it a trivial corollary because `IsHardlyRamified` already says that the det is cyclo?

FLT/GaloisRepresentation/HardlyRamified/Family.lean
kbuzzard: Yes I think so. If M is module-finite over Z_p then by the structure theorem M = Z_p^d x prod_{i=1}^n Z/p^{m_i} Z and the module topology is the product topology, which is the p-adic topology. So it suffices to show that p is in m and that some power of m is in (p). First let's prove that p is in m. If R is a local Z_p-algebra, module-finite over Z_p, then R/m is a field module-finite over Z_p, so is either finite (and thus p \in m and we're done) or contains Q_p (which can't happen because a Noetherian Z_p-module can't contain a non-Noetherian Z_p-submodule). Hence p is in m. 

For the converse I need to check that some power of m is in (p) or equivalently that some power of m is 0 in R/pR but R/pR is Artin local so m is nilpotent by some Nakayama argument. Do you believe me?

FLT/GaloisRepresentation/HardlyRamified/Family.lean
kbuzzard: Oh nice question. I had it in my head that A, being an integral domain module-finite and free over Z_ell is necessarily an order in a finite extension of Q_ell (namely its field of fractions) and this *probably* gives the scalar tower result but I don't see any harm in assuming it. As for the topologies I had it in my head that of course A would have the subspace topology coming from the local field topology but I don't see any reason why that would be true, e.g. right now I think A can just be Z_ell with the discrete topology, so this is wrong. Thanks!

kbuzzard: I've made A have the Z_ell-module topology, this is definitely the right topology and it would not surprise me if it was now a fun exercise to fill in all your other suggestions.

FLT/GaloisRepresentation/HardlyRamified/Family.lean
kbuzzard: lol oops thanks!

FLT/GaloisRepresentation/HardlyRamified/ModThree.lean
kbuzzard: Yeah exactly. A lot of what Taylor does can be golfed in this way. 

One thing that occurred to me was that in the general case (e.g. R (compact Hausdorff) local, where I am also using this trick) one has to prove that the kernel of a surjection R^2->R is isomorphic to R. This is OK because if (1,0) goes to a and (0,1) goes to b then the ideal (a,b) is all of R so at least one of a and b is not in the maximal ideal and is thus a unit, meaning that you can change basis to make the map the projection onto the first coordinate.

FLT/GaloisRepresentation/HardlyRamified/Lift.lean
```diff
@@ -0,0 +1,32 @@
...
+/--
+An irreducible mod p hardly ramified represntation lifts to a p-adic one.
+-/
+theorem lifts (ρ : GaloisRep ℚ k V) (hρirred : ρ.IsIrreducible)
+    (hρ : IsHardlyRamified hpodd hV ρ) :
+    ∃ (R : Type u) (_ : CommRing R) (_ : IsLocalRing R)
+      (_ : TopologicalSpace R) (_ : IsTopologicalRing R)
+      (_ : Algebra ℤ_[p] R) (_ : IsLocalHom (algebraMap ℤ_[p] R))
+      (_ : Module.Finite ℤ_[p] R) (_ : Module.Free ℤ_[p] R)
+      (_ : Algebra R k) (_ : IsScalarTower ℤ_[p] R k) (_ : ContinuousSMul R k)
```
erdOne, line 27: I think you also want module topology somewhere here?

kbuzzard: Heh, I was thinking "oh didn't I prove that the topology had to be the module topology somewhere?" but this is nonsense, what I proved was that if R had the module topology then it was automatically a topological ring (if it's a module-finite Z_p-algebra). Thanks!

PR #744, opened 2025-10-10
Bijection of lTensor

### LINKED ISSUE #243: Proof that L (x)[K] A_K^f = A_L^f

maddycrim:
Closes #243 

Uses #697.

### DIFF
--- a/FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
+++ b/FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
@@ -38,6 +38,11 @@ variable {A : Type*} [CommRing A] {ι : Type*} {R : ι → Type*} {ℱ : Filter 
 
 open Set Filter RestrictedProduct
 
+/-- A linear map version of `RestrictedProduct.inclusion` :
+if `𝓕 ≤ 𝓖` then there's a linear map
+`Πʳ i, [R i, C i]_[𝓖] →ₗ[A] Πʳ i, [R i, C i]_[𝓕]` where the `R i`
+are `A`-modules and the `C i` are submodules.
+-/
 def inclusionLinearMap {𝓕 𝓖 : Filter ι} (h : 𝓕 ≤ 𝓖) :
     Πʳ i, [R i, C i]_[𝓖] →ₗ[A] Πʳ i, [R i, C i]_[𝓕] :=
   mapAlongLinearMap R R id h (fun _ ↦ .id)
@@ -161,3 +166,4 @@ instance instIsDirectLimit' : IsDirectLimit (M := fun (S : ℱ.setsᵒᵈ) ↦ �
 end module
 
 end RestrictedProduct
+#lint


FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,156 @@
...
+variable [IsDirectLimit M P f g]
+
+instance isDirectLimit : IsDirectLimit M (Module.DirectLimit M f) f
+  (Module.DirectLimit.of R ι M f) where
+  surj := Module.DirectLimit.exists_of
+  inj i j mi mj h := by
+    apply_fun Module.DirectLimit.linearEquiv _ _ at h
+    simp_rw [Module.DirectLimit.linearEquiv_of] at h
+    have ⟨k, hi, hj, hij⟩ := Quotient.exact h
+    exact ⟨k, hi, hj, hij⟩
```
kbuzzard: Snap! So this means `exact Quotient.exact h` works instead of these two lines, which means that `simpa [Module.DirectLimit.linearEquiv_of] using h` closes the goal already on line 40.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,156 @@
...
+variable (P₁ P₂ : Type*) [AddCommMonoid P₁] [Module R P₁] [AddCommMonoid P₂] [Module R P₂]
+  (g₁ : ∀ i, M i →ₗ[R] P₁) (g₂ : ∀ i, M i →ₗ[R] P₂)
+
+open Classical in
+/--
+The universal property of the direct limit: define a linear map from the direct limit
+by giving a compatible family of linear maps from the components.
+Each element of the limit is sent according to its representative in some component.
+-/
+noncomputable def lift [IsDirectLimit M P₁ f g₁] (Hg : ∀ i j hij x, g₂ j (f i j hij x) = g₂ i x) :
```
kbuzzard: The full name of this def is now called `Module.lift` which is probably a bad name (because you would not expect something with that name to be about `IsDirectLimit`) . I think that almost immediately after making the definition `IsDirectLimit` you want to `namespace IsDirectLimit` and put everything about the class in that namespace. This would be the standard approach. In general it's a useful tip to think "what other definitions already in mathlib are like this" and then to take a look at how they are set up, and mimic the approach. For example `IsDirectLimit` is formally quite similar to `IsLocalization` which is already in mathlib [here](https://github.com/leanprover-community/mathlib4/blob/24899e1184118f6ac3a1584767be5e6ad9c8d18e/Mathlib/RingTheory/Localization/Defs.lean#L97) and you can see that this is what they do.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,156 @@
...
+import Mathlib.Algebra.Colimit.Module
+
+section IsDirectLimit
+
+namespace Module
+
+variable {R : Type*} [Semiring R] {ι : Type*} [Preorder ι] (M : ι → Type*) (P : Type*)
+ [AddCommMonoid P] [Module R P] [∀ i, AddCommMonoid (M i)] [∀ i, Module R (M i)]
+ (f : ∀ i j, i ≤ j → M i →ₗ[R] M j) (g : ∀ i, M i →ₗ[R] P) [DecidableEq ι] [IsDirected ι (· ≤ ·)]
+ [Nonempty ι] [DirectedSystem M (f · · ·)]
```
kbuzzard: You put the hypotheses `[DecidableEq ι] [IsDirected ι (· ≤ ·)] [Nonempty ι] [DirectedSystem M (f · · ·)]` as variables here, but none of them are needed in the *definition* of `IsDirectLimit` so none of them are included in it (you can see this by hovering over `IsDirectLimit` and seeing that none of them are mentioned). And if you think about it, you can see the logic: the definition of `IsDirectLimit` makes sense without any of those assumptions
(it's just unlikely to be useful). I am always in two minds about what to do with situations like this but ultimately my impression is that people like Lean code to work like this, with definitions as weak as possible, so I don't think anything needs changing here, I just thought I'd flag it.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,156 @@
...
+      have hx := Classical.choose_spec (Classical.choose_spec (IsDirectLimit.surj f (g:= g₁) x))
+      have hy := Classical.choose_spec (Classical.choose_spec (IsDirectLimit.surj f (g:= g₁) y))
+      have hxy := Classical.choose_spec (Classical.choose_spec
+        (IsDirectLimit.surj f (g:= g₁) (x + y)))
+      set ix := Classical.choose (IsDirectLimit.surj f (g:= g₁) x)
+      set mx := Classical.choose (Classical.choose_spec (IsDirectLimit.surj f (g:= g₁) x))
+      set iy := Classical.choose (IsDirectLimit.surj f (g:= g₁) y)
+      set my := Classical.choose (Classical.choose_spec (IsDirectLimit.surj f (g:= g₁) y))
+      set ixy := Classical.choose (IsDirectLimit.surj f (g:= g₁) (x + y))
+      set mxy := Classical.choose (Classical.choose_spec (IsDirectLimit.surj f (g:= g₁) (x + y)))
```
kbuzzard: Yeah this is not good. I mean it *works*, and great job making it work! But now stepping back, I think it might be evidence that directly after the definition of IsDirectLimit and in the IsDirectLimit namespace you want some basic API. Here is a suggestion:

```lean
@[mk_iff] class IsDirectLimit : Prop where
  surj : ∀ m : P, ∃ i, ∃ mi : M i, g i mi = m
  inj :  ∀ i j, ∀ mi : M i, ∀ mj : M j, g i mi = g j mj → ∃ (k : ι) (hik : i ≤ k) (hjk : j ≤ k),
      f i k hik mi = f j k hjk mj
  compatibility : ∀ i j hij x, g j (f i j hij x) = g i x

variable [IsDirectLimit M P f g]

namespace IsDirectLimit

variable {M P}

theorem is_surjective [IsDirectLimit M P f g] : ∀ m : P, ∃ i, ∃ mi : M i, g i mi = m :=
  IsDirectLimit.surj f

noncomputable def preimage_index [IsDirectLimit M P f g] (p : P) : ι := (is_surjective f g p).choose

noncomputable def preimage [IsDirectLimit M P f g] (p : P) : M (preimage_index f g p) :=
  (is_surjective f g p).choose_spec.choose

theorem image_preimage [IsDirectLimit M P f g] (p : P) :
    g (preimage_index f g p) (preimage f g p) = p :=
   (is_surjective f g p).choose_spec.choose_spec

```

First of all, it's really annoying that `surj` has g implicit. An algorithm decided that you wanted this but your code is clearly saying that you didn't. This is why so many structure fields in mathlib are called `blah'` instead of `blah`; the algorithm made a decision which is inconvenient in practice and so the human has to fix it. Instead of priming `surj` the first thing I do is define `is_surjective` which has the convenient implicit/explicit choices (I make M and P implicit, you need them explicit for the definition of `IsDirectLimit` but probably for the theorems you want them implicit). Now have the right surjective, I write the API need for all this Classical.choose (Classical.choose_spec) stuff *once*. Now you don't need any of those haves and sets because you can just do stuff like
```lean 
      obtain ⟨k, hxk, hyk⟩ := 
        IsDirected.directed (r := (· ≤ ·)) (preimage_index f g₁ x) (preimage_index f g₁ y)
```

Note also that in my theorems a linter at my end is going nuts -- saying that there's a ton of hypotheses which are not needed in the theorems. This is because you've put the final few `variable`s around line 16 too early; those not needed for the definition are also not needed for these basic theorems. You should probably move these later.

erdOne: You can also do `obtain ⟨k, hxk, hyk⟩ := IsDirectLimit.is_surjective f (g := g₁) x` in line instead of introducing the  `preimage_index` defs that Kevin is mentioning. Hence I'm not sure if we actually want the`preimage_index` defs or not.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,226 @@
...
+section RestrictedProduct
+
+open Set Filter
+
+variable {ι : Type*} {𝓕 : Filter ι}
+
+variable (𝓕) in
+/-- The complements of sets in a `Filter`.
+E.g. for the cofinite filter, these are just the finite subsets. -/
+def Filter.complement : Set (Set ι) := (fun S => Sᶜ) '' 𝓕.sets
```
erdOne: I am slightly not sure if we want this or not. I am guessing we would prefer passing `hS : S ∉ 𝓕` around.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
erdOne: I think we should require `[DirectedSystem M (f · · ·)]` in `IsDirectLimit`.

erdOne (commented):
This is nice and we definitely want something similar in mathlib in the near future.
Here are some passing comments:

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
maddycrim: Fixed!

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
maddycrim: I've changed the code so that everything after `IsDirectLimit` is in namespace `IsDirectLimit`.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
maddycrim: I went ahead and got rid of the unused hypotheses and added a `directedsystem` condition to `IsDirectLimit`.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
maddycrim: I went ahead and added this along with theorems for injectivity and compatibility. I did keep the `preimage` and `preimage_index` - it makes defining the map a bit cleaner. Also, when proving `map_add`, using `obtain ⟨k, hxk, hyk⟩ := IsDirectLimit.is_surjective f (g := g₁) x` seems to give additional steps. I don't think it's automatic that the index we get from `obtain` is the same as the one we get from `preimage_index`. We could use the injectivity on the equality `g₁ (preimage_index f g₁ x) (preimage f g₁ x) = g₁ k hxk`, but I at least don't immediately see the benefit of this route.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
maddycrim: If we pass `hS : S ∉ 𝓕` instead, I'm not sure how to see `filter_bot` works which is needed for `inclusion_to_restrictedProduct`. I don't think it's necessarily true that if `S \nin \MCF` that `S\^c in \MCF` unless we have an ultrafilter.

kbuzzard: Yes you're right, this isn't `S ∉ 𝓕`, this is `Sᶜ ∈ 𝓕`, it's a so-called ideal. We seem to have them in mathlib as `Order.Ideal` (defined on general preorders rather than just on `Set X` which is what you want here) but I can't see the link to filters.

erdOne: Yes you are right it is `Sᶜ ∈ 𝓕` instead. But I think my suggestion still stands as I think we would prefer passing `Sᶜ ∈ 𝓕` around.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,229 @@
...
+    · unfold c
+      rw [compl_inter, hsi.2, hsj.2]
+  use ⟨c, this⟩
+  constructor <;>
+  simp [c]
+
+instance RestrictedProductIsDirectLimit :
+  IsDirectLimit (mem_A_away_from_S C)
+  Πʳ i, [R i, C i]_[ℱ] (inclusion_module (ℱ := ℱ))
+  (inclusion_to_restricted_product_module ) where
```
erdOne: I think this works in the generality where you have a directed family of filters whose intersection is `ℱ` right? Perhaps it would also be easier to prove this way.

erdOne: So something like this perhaps
```lean
variable {A : Type*} [CommRing A] {R : ι → Type*} {ℱ : Filter ι}
  [Π i, AddCommGroup (R i)] [∀ i, Module A (R i)] {C : ∀ i, Submodule A (R i)}
  {I : Type*} [Preorder I] (𝓖 : I → Filter ι) (h𝓖 : Antitone 𝓖) (h𝓕 : 𝓕 = iInf 𝓖)

/-- If `𝓕 ≤ 𝓖`, the restricted product `Πʳ i, [R i, A i]_[𝓖]` is naturally included in
`Πʳ i, [R i, A i]_[𝓕]`. This is the corresponding map. -/
def RestrictedProduct.linclusion
    {ι : Type*} {R₀ : Type*} (R : ι → Type*) [Semiring R₀] [∀ i, AddCommMonoid (R i)]
    [∀ i, Module R₀ (R i)] (A : (i : ι) → Submodule R₀ (R i)) {𝓕 𝓖 : Filter ι}
    (h : 𝓕 ≤ 𝓖) : Πʳ i, [R i, A i]_[𝓖] →ₗ[R₀] Πʳ i, [R i, A i]_[𝓕] where
  toFun := inclusion R (A ·) h
  map_add' _ _ := rfl
  map_smul' _ _ := rfl

lemma RestrictedProduct.isDirectLimit :
    IsDirectLimit (Πʳ i, [R i, C i]_[𝓖 ·]) Πʳ i, [R i, C i]_[𝓕]
    (linclusion _ _ <| @h𝓖 · · ·) (linclusion _ _ <| h𝓕.trans_le <| iInf_le 𝓖 ·) := sorry
```
Edit: I probably also need `I` is directed.

erdOne (changes requested):

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,151 @@
...
+/--
+`IsDirectLimit f g` states that `P` is the direct limit of the system `(M i, f i j hij)`.
+It requires:
+* `surj`: says every element of `P` comes from some `M i` via the canonical map `g i`.
+* `inj`: says that if two elements of (possibly different) modules `M i` and `M j` map
+  to the same element of `P`, then they become equal after being mapped into a common
+  later module `M k`.
+* `compatibility`: ensures that the maps `g i` and `f i j hij` are compatible with one another.
+-/
+@[mk_iff] class IsDirectLimit : Prop where
```
erdOne: Also I think we should state this in terms of the underlying function because the direct limit is the same as the direct limit of the underlying sets, and `DirectedSystem` is also stated in terms of unbundled functions.
```
variable {ι : Type*} [Preorder ι] (M : ι → Type*) (P : Type*)
 (f : ∀ i j, i ≤ j → M i → M j) (g : ∀ i, M i → P) in
@[mk_iff] class IsDirectLimit [DirectedSystem M f] : Prop where
  surj : ∀ m : P, ∃ i, ∃ mi : M i, g i mi = m
  inj :  ∀ i j, ∀ mi : M i, ∀ mj : M j, g i mi = g j mj → ∃ (k : ι) (hik : i ≤ k) (hjk : j ≤ k),
      f i k hik mi = f j k hjk mj
  compatibility : ∀ i j hij x, g j (f i j hij x) = g i x
```

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
kbuzzard: I am working on removing this definition and more generally this file from the PR.

FLT/Mathlib/Algebra/IsDirectLimit.lean
```diff
@@ -0,0 +1,188 @@
+/-
```
kbuzzard, line 1: This file should definitely not be in this directory; it should probably be in `FLT/Mathlb/Algebra/Colimit/Module.lean` which means "stuff which should go in `Mathlib.Algebra.Colimit.Module` but we didn't put it there yet".

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimit.lean
```diff
@@ -0,0 +1,188 @@
...
+    simp_rw [Module.DirectLimit.linearEquiv_of] at h
+    simpa [Module.DirectLimit.linearEquiv_of] using h
+  compatibility i j hij x := Module.DirectLimit.of_f
+
+variable [AddCommMonoid P₁] [Module R P₁] [AddCommMonoid P₂] [Module R P₂]
+  (g₁ : ∀ i, M i →ₗ[R] P₁) (g₂ : ∀ i, M i →ₗ[R] P₂) [h₁ : IsDirectLimit M P₁ (f · · ·) (g₁ · ·)]
+
+omit [IsDirected ι (· ≤ ·)] in
+theorem compatibility_module {i j hij} (x : M i) : g₁ j (f i j hij x) = g₁ i x :=
+  compatibility' (f := (f · · ·)) (g := (g₁ · ·)) x
```
kbuzzard:
```suggestion
  compatibility' (f · · ·) (g₁ · ·) x
```
f and g are explicit inputs, you don't need (f := ...). Same comment for line 143.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,197 @@
...
+  intro S
+  simp only [le_principal_iff]
+  exact S.2
+
+open scoped RestrictedProduct
+
+variable {R : ι → Type*} {A : ι → Type*} [Π i, SetLike (A i) (R i)] {C : Π i, A i}
+
+variable (C) in
+/-- This is (isomorphic to) `(Π i ∈ S, R i) × (Π i ∉ S, A i)` -/
```
kbuzzard:
```suggestion
/-- This is (isomorphic to) `(Π i ∈ S, C i) × (Π i ∉ S, R i)` -/
```

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+    (S : Set ι) [∀ i, Decidable (i ∈ S)] :
+   (Πʳ i, [R i, C i]_[𝓟 S]) ≃ₗ[A] ((Π i : {i // i ∈ S}, C i) ×
+  (Π i : {i // i ∉ S}, R i)) where
+    __ := principalAddEquivSum R S (A := C)
+    map_smul' m x := by
+      simp only [AddEquiv.toEquiv_eq_coe, Equiv.toFun_as_coe, EquivLike.coe_coe, RingHom.id_apply]
+      dsimp [principalAddEquivSum, principalEquivProd]
+      ext i
+      · simp only [Pi.smul_apply, SetLike.coe_sort_coe, SetLike.val_smul]
+      · simp only [Pi.smul_apply]
```
kbuzzard:
```suggestion
    map_smul' _ _ := rfl
```

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+    (linclusion _ _ <| @h𝓖 · · ·) (linclusion _ _ <| hℱ.trans_le <| iInf_le 𝓖 ·) where
+  inj Sᵢ Sⱼ mi mj hmij := by
+    obtain ⟨Sₖ, hik, hjk⟩ := @directed_of _ (· ≤ ·) _ Sᵢ Sⱼ
+    refine ⟨Sₖ, hik, hjk, ?_⟩
+    dsimp [RestrictedProduct.inclusion, RestrictedProduct.inclusion]
+    dsimp [coeLinearMap,
+      RestrictedProduct.inclusion] at hmij
+    apply Subtype.ext
+    simp only
+    injection hmij
```
kbuzzard, line 144:
```suggestion
    refine ⟨Sₖ, hik, hjk, ?_⟩
    apply Subtype.ext
    injection hmij
```
These calls to the simplifier are only changing things up to definitional equality and are not needed. Same for the surj and compatibility proofs.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+  compatibility i j hij x := by
+    dsimp [coeLinearMap, RestrictedProduct.inclusion, RestrictedProduct.inclusion]
+    exact Subtype.ext rfl
+
+instance instIsDirectLimit' : IsDirectLimit (fun (S : ℱ.setsᵒᵈ) ↦ Πʳ i, [R i, C i]_[𝓟 S.1])
+    (Πʳ i, [R i, C i]_[ℱ]) (inclusionLinearMap · · ·) (coeLinearMap ·) := by
+  apply instIsDirectLimit
+  · intro i j hij
+    simp only [le_principal_iff, mem_principal]
+    exact hij
```
kbuzzard:
```suggestion
    simpa only [le_principal_iff, mem_principal]
```
`simpa` is basically `simp` + `assumption`

FLT/DedekindDomain/FiniteAdeleRing/TensorRestrictedProduct.lean
```diff
@@ -108,10 +108,74 @@ lemma lTensorPrincipalEquiv_tmul (m : M) (x : Πʳ i, [N i, L i]_[𝓟 S]) (i :
...
-lemma lTensor_bijective : Function.Bijective (lTensor R M N ℱ L) :=
-  -- Should follow from the above and general results about direct
-  -- limits of tensor products.
-  sorry
+open scoped Filter in
+lemma lTensor_bijective : Function.Bijective (lTensor R M N ℱ L) := by
+  classical
+  let comp1 := TensorProduct.directLimitRight (inclusionLinearMap (C:=L) (ℱ:=ℱ)) M (R:=R)
+  let comp := IsDirectLimit.Module.linearEquiv
+     (Πʳ i, [N i, L i]_[ℱ])
```
kbuzzard:
```suggestion
  let comp := IsDirectLimit.Module.linearEquiv (Πʳ i, [N i, L i]_[ℱ])
```

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+open scoped RestrictedProduct TensorProduct IsDirectLimit
+
+variable {A : Type*} [CommRing A] {ι : Type*} {R : ι → Type*} {ℱ : Filter ι}
+  [Π i, AddCommGroup (R i)] [∀ i, Module A (R i)] {C : ∀ i, Submodule A (R i)} {M : Type*}
+  [AddCommGroup M] [Module A M] [Module.FinitePresentation A M] (S : ℱ.setsᵒᵈ)
+
+open Set Filter RestrictedProduct
+
+/-- Linear map version of `inclusion`. -/
+def inclusionLinearMap (S₁ S₂ : ℱ.setsᵒᵈ) (h : S₁ ≤ S₂) :
```
kbuzzard:
```suggestion
def inclusionLinearMap {S₁ S₂ : ℱ.setsᵒᵈ} (h : S₁ ≤ S₂) :
```
I should think this will be more efficient?

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+`(Π i ∈ S, R i) × (Π i ∉ S, A i)`
+-/
+def principalEquivProd : Πʳ i, [R i, A i]_[𝓟 S] ≃
+    (Π i : {i // i ∈ S}, A i) × (Π i : {i // i ∉ S}, R i) where
+  toFun x := (fun i ↦ ⟨x i, x.property i.property⟩, fun i ↦ x i)
+  invFun y := ⟨fun i ↦ if hi : i ∈ S then y.1 ⟨i, hi⟩ else y.2 ⟨i, hi⟩,
+  Filter.eventually_principal.mpr (fun i hi ↦ by simp only [hi]; exact (y.1 ⟨i, hi⟩).2)⟩
+  left_inv x := by
+    simp only [dite_eq_ite, ite_self]
+    rfl
```
kbuzzard:
```suggestion
  left_inv x := by
    ext
    simp
```
I think it's cleaner like that. Try to avoid explicit lemmas if you can. Tactic proofs are more robust.

FLT/Mathlib/Algebra/IsDirectLimit.lean
```diff
@@ -0,0 +1,188 @@
...
+Copyright (c) 2025 Madison Crim. All rights reserved.
+Released under Apache 2.0 license as described in the file LICENSE.
+Authors: Madison Crim
+-/
+
+import Mathlib.Algebra.Colimit.Module
+
+variable {ι : Type*} [Preorder ι] (M : ι → Type*) (P : Type*)
```
kbuzzard:
```suggestion
variable {ι : Type*} [Preorder ι] {M : ι → Type*{ {P : Type*}
```
I tried the refactor and I like it. It saves a bunch of characters. Most of the fixes just involve deleting M and P; I have mentioned the two cases where this is not the case.

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+
+instance {I : Type*} [Preorder I] (𝓖 : I → Filter ι) (h𝓖 : Antitone 𝓖) :
+    DirectedSystem (fun x ↦ Πʳ (i : ι), [R i, ↑(C i)]_[𝓖 x]) (linclusion _ _ <| @h𝓖 · · ·) where
+  map_self _ _ := rfl
+  map_map _ _ _ _ _ _ := rfl
+
+instance instIsDirectLimit {I : Type*} [Preorder I] [Nonempty I] [IsDirected I (· ≤ ·)]
+    (𝓖 : I → Filter ι) (h𝓖 : Antitone 𝓖) (hℱ : ℱ = iInf 𝓖) :
+    IsDirectLimit (Πʳ i, [R i, C i]_[𝓖 ·]) Πʳ i, [R i, C i]_[ℱ]
+    (linclusion _ _ <| @h𝓖 · · ·) (linclusion _ _ <| hℱ.trans_le <| iInf_le 𝓖 ·) where
```
kbuzzard: If you make M and P implicit in the definition of `IsDirectLimit` (and I think you should, having tried it myself) then this becomes
```lean
    IsDirectLimit (linclusion R C <| @h𝓖 · · ·) (linclusion R C <| hℱ.trans_le <| iInf_le 𝓖 ·) where
```
(you can remove all the products and all you have to pay is filling in the underscores)

FLT/DedekindDomain/FiniteAdeleRing/IsDirectLimitRestricted.lean
```diff
@@ -0,0 +1,182 @@
...
+    use j
+    apply RestrictedProduct.exists_inclusion_eq_of_eventually _ _ _ hj
+    rw [hℱ]
+    exact iInf_le_iff.mpr fun b a ↦ a j
+  compatibility i j hij x := by
+    dsimp [coeLinearMap, RestrictedProduct.inclusion, RestrictedProduct.inclusion]
+    exact Subtype.ext rfl
+
+instance instIsDirectLimit' : IsDirectLimit (fun (S : ℱ.setsᵒᵈ) ↦ Πʳ i, [R i, C i]_[𝓟 S.1])
+    (Πʳ i, [R i, C i]_[ℱ]) (inclusionLinearMap · · ·) (coeLinearMap ·) := by
```
kbuzzard: If you make M and P implicit then this becomes
```
instance instIsDirectLimit' : IsDirectLimit
    (M := fun (S : ℱ.setsᵒᵈ) ↦ Πʳ i, [R i, C i]_[𝓟 S.1])
    (inclusionLinearMap · · ·) (coeLinearMap ·) := by
```
but still this is shorter than what you are having to write.

PR #753, opened 2025-10-20
chore: change linter options in lakefile

kbuzzard:
An attempt to silence the warnings I get with a scratch file.

See [#general > weak.linter.mathlibStandardSet and lint-style-action @ 💬](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/weak.2Elinter.2EmathlibStandardSet.20and.20lint-style-action/near/545873307)


grunweg:
As I said on zulip ([#general > weak.linter.mathlibStandardSet and lint-style-action @ 💬](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/weak.2Elinter.2EmathlibStandardSet.20and.20lint-style-action/near/545906199)), I think this is not the right fix. Instead, your scratch file should import something from mathlib (such as `Mathlib/Init.lean`) instead.
And yes, there should perhaps be a better error about this, suggesting this fix.

PR #760, opened 2025-10-30
Resolve remaining `sorry`s in `AdeleRing.lean`

smmercuri:
Resolves three sorries `finiteAdeleRing_equiv_qHat`, `principalSubgroup_equiv_ratsub`, `finiteIntegralAdeles_equiv_zHatsub` in `AdeleRing.lean` by the following approach:
1. In the pending mathlib PR #30576 there is the sorry-free result that `𝔭.adicCompletion ℚ ≃ₐ[ℚ] ℚ_[𝔭]` and this restricts to the ring of integers. 
2. Lift this isomorphism of 1. to restricted products `FiniteAdeleRing (𝓞 ℚ) ℚ ≃ₐ[ℚ] Πʳ (p : Nat.Primes), [ℚ_[p], subring p]`
3. In the `Padic` case prove directly that `∃ q : ℚ, ∀ p : Nat.Primes, q - x p ∈ subring p` using CRT and the reduction maps `ℤ_[p] →+* ZMod (p ^ n)` which already exist in mathlib, and use the isomorphism in 2. to lift back to the rational finite adele ring.

Additionally:
- Move equivalence results on restricted products from `RestrictedProduct.Basic` to a new file `RestrictedProduct.Equiv`
- Move `RestrictedProduct.structureMapMonoidHom` from `RestrictedProduct.Topology` to `RestrictedProduct.Basic` and expand to `RingHom` versions.
- Remove `RestrictedProduct.congrRight` since this is `RestrictedProduct.map` from mathlib.
- Do some `#min_imports`
- Tidy up `AdeleRing.lean` slightly

### DIFF
--- a/FLT/Mathlib/Data/Nat/Prime/Defs.lean
+++ b/FLT/Mathlib/Data/Nat/Prime/Defs.lean
@@ -1,7 +1,0 @@
-import Mathlib.Data.Nat.Prime.Defs
-
-namespace Nat.Primes
-
-instance {p : Nat.Primes} : Fact p.1.Prime := ⟨p.2⟩
-
-end Nat.Primes
--- a/FLT/NumberField/Padics/RestrictedProduct.lean
+++ b/FLT/NumberField/Padics/RestrictedProduct.lean
@@ -1,5 +1,4 @@
 import FLT.DedekindDomain.FiniteAdeleRing.BaseChange
-import FLT.Mathlib.Data.Nat.Prime.Defs
 import Mathlib.NumberTheory.Padics.RingHoms
 
 open IsDedekindDomain NumberField PadicInt RestrictedProduct
@@ -55,6 +54,8 @@ end IsDedekindDomain.HeightOneSpectrum
 
 namespace Padic
 
+local instance {p : Nat.Primes} : Fact p.1.Prime := ⟨p.2⟩
+
 theorem setOf_norm_one_lt_finite (x : ℚ) :
     {p : Nat.Primes | 1 < ‖(x : ℚ_[p])‖}.Finite := by
   apply Set.Finite.subset _ fun p ↦ mt (Padic.norm_rat_le_one) ∘ not_le.2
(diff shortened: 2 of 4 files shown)


kbuzzard:
Oh sorry, I just bumped mathlib and it's caused a conflict: I changed a `simp [c]` to `simp [-algebraMap.coe_inj, c]` to fix a proof. Having caused you trouble I'll now review mathlib #30576 (edit: actually no, I'll review the PRs it depends on)

FLT/Mathlib/Data/Nat/Prime/Defs.lean
```diff
@@ -0,0 +1,7 @@
+import Mathlib.Data.Nat.Prime.Defs
+
+namespace Nat.Primes
+
+instance {p : Nat.Primes} : Fact p.1.Prime := ⟨p.2⟩
```
kbuzzard: My instinct is that there should be no global `Fact` instances anywhere, they are a slowdown hazard. Can you give this instance a name, make it a def, and then make it a local instance in the files which need it?

FLT/NumberField/Padics/RestrictedProduct.lean
```diff
@@ -0,0 +1,175 @@
...
+
+namespace RestrictedProduct
+
+theorem padic_valuation_neg_of_mem_indexSupport
+    {x : Πʳ (p : Nat.Primes), [ℚ_[p], subring p]} {p : Nat.Primes} (hp : p ∈ x.indexSupport) :
+    (x p).valuation < 0 := by
+  contrapose! hp
+  simpa [Padic.norm_le_one_iff_val_nonneg]
+
+/-- The natural `n : ℕ` for which `x p * n` is a `p`-adic integer for all `p`. -/
```
kbuzzard:
```suggestion
/-- The smallest positive natural `n : ℕ` for which `x p * n` is a `p`-adic integer for all `p`. -/
```

FLT/Mathlib/Data/Nat/Prime/Defs.lean
smmercuri: Made this a local instance where needed

PR #761, opened 2025-11-05
feat: prove FreyCurve.torsion_not_isIrreducible modulo 1980s (#760)

stepan2698-cpu:
This PR reduces the proof of FreyCurve.torsion_not_isIrreducible modulo other hard theorems in other files to the fact that a Galois representation with characteristic polynomial of Frobenius at p (X-1)(X-p) is reducible, which was known in the 1980s.


FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
 it is defined in a completely nonconstructive way, so we add the classical instance. -/
 noncomputable local instance : DecidableEq (AlgebraicClosure ℚ) := Classical.typeDecidableEq _
 
+section GaloisRep_API
```
kbuzzard: This section should be in the file `FLT/Deformations/RepresentationTheory/GaloisRep.lean`

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
+
+variable {K : Type*} {L : Type*} [Field K] [Field L]
+variable {A : Type*} [CommRing A] [TopologicalSpace A]
+variable {B : Type*} [CommRing B] [TopologicalSpace B] [IsTopologicalRing B]
+variable [Algebra A B] [ContinuousSMul A B]
+variable {M N : Type*} [AddCommGroup M] [Module A M] [AddCommGroup N] [Module A N]
+variable [Module.Finite A M] [Module.Free A M] [Module.Finite A N] [Module.Free A N]
+
+lemma charpoly_base_change (ρ : GaloisRep K A M) (g : Field.absoluteGaloisGroup K) :
```
kbuzzard: This (and the other declarations in this section) should be in the `GaloisRep` namespace (root namespace) and should be called `charpoly_baseChange` (the full list of naming convention rules are at https://leanprover-community.github.io/contribute/naming.html ).

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
+  LinearMap.charpoly (GaloisRep.baseChange B ρ g) =
+    Polynomial.map Algebra.algebraMap (LinearMap.charpoly (ρ g)) := by
+  apply LinearMap.charpoly_baseChange
+
+lemma charpoly_conj (ρ : GaloisRep K A M) (e : M ≃ₗ[A] N) (g : Field.absoluteGaloisGroup K) :
+  LinearMap.charpoly ((GaloisRep.conj ρ e) g) = LinearMap.charpoly (ρ g) := by
+  apply LinearEquiv.charpoly_conj
+
+omit [Module.Finite A M] [Module.Free A M] [Module.Finite A N] [Module.Free A N] in
+lemma conj_commutes_with_to_local [NumberField K]
```
kbuzzard: This can just be `GaloisRep.conj_toLocal` for now (in general it's fine to name a lemma with only part of what is going on in the statement as long as there's no competing lemma wanting to use the same name). Similarly the next one can be `GaloisRep.baseChange_toLocal`.

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
+  rfl
+
+lemma to_local_commutes_with_base_change [NumberField K]
+   (v : IsDedekindDomain.HeightOneSpectrum (NumberField.RingOfIntegers K)) (ρ : GaloisRep K A M) :
+    (GaloisRep.baseChange B ρ).toLocal v = GaloisRep.baseChange B (ρ.toLocal v) := by
+  rfl
+
+end GaloisRep_API
+
+lemma polynomial_map_trans (A B C : Type*) [CommRing A] [CommRing B] [CommRing C]
```
kbuzzard: `exact?` informs me that this already exists as `Polynomial.map_map`

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
+  (f : A →+* B) (g : B →+* C) (P : Polynomial A) :
+    Polynomial.map g (Polynomial.map f P) = Polynomial.map (g.comp f) P := by
+  change (⇑(Polynomial.mapRingHom g) ∘ ⇑(Polynomial.mapRingHom f)) P =
+    ⇑(Polynomial.mapRingHom (g.comp f)) P
+  rw[← RingHom.coe_comp, Polynomial.mapRingHom_comp]
+
+lemma ideal_in_integral_extension_overlap (R S : Type*) [CommRing R] [Nontrivial R] [CommRing S]
```
kbuzzard: The conclusion of this lemma should be stated as `I.comap (algebraMap R S) ≠ ⊥` (which would be more idiomatic, especially as a hypothesis is of this form), and be called `Ideal.comap_ne_bot` or perhaps `Ideal.comap_ne_bot_of_isIntegral` if there are other reasons why the preimage of a nonzero ideal is nonzero.

Hmm...with this in mind, searching for `comap_ne_bot` lemmas gives me `Ideal.comap_ne_bot_of_algebraic_mem` which is actually stronger because it assumes only that an element is algebraic rather than assuming that the whole extension is integral, although like your statement it's asymmetric.  So maybe we want (note also that I made `S` and `I` implicit, because they can both be inferred from the type of the later input `hI`)

```
lemma Ideal.comap_ne_bot_of_isIntegral (R : Type*) {S : Type*} [CommRing R] [Nontrivial R]
    [CommRing S] [IsDomain S] [Algebra R S] [Algebra.IsIntegral R S] {I : Ideal S} (hI : I ≠ ⊥) :
    I.comap (algebraMap R S) ≠ ⊥ := by
  obtain ⟨i, hiI, hi0⟩ := Submodule.exists_mem_ne_zero_of_ne_bot hI
  exact Ideal.comap_ne_bot_of_algebraic_mem hi0 hiI <| Algebra.IsAlgebraic.isAlgebraic i
```
(this could go in `FLT/Mathlib/RingTheory/Ideal/GoingUp.lean` for now, but really you should make a PR to mathlib in the corresponding going-up file, adding it just after `Ideal.comap_ne_bot_of_algebraic_mem` and then add a note in the FLT file referring to the mathlib PR) and then the statement in this PR can be proved without too much difficulty from it (use `Submodule.ne_bot_iff` again), but we probably don't want it at all and could get away with `Ideal.comap_ne_bot_of_isIntegral` in applications.

Note: I'm sure that working through your proof was great practice for doing commutative algebra in Lean! This is not time wasted.

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
+    rw[← neg_eq_iff_add_eq_zero] at hP
+    rw[← hP, Submodule.neg_mem_iff]
+    exact ⟨hQ2, by exact
+      Ideal.IsTwoSided.mul_mem_of_left (Polynomial.eval₂ (algebraMap R S) a Q') ha⟩
+
+theorem galois_rep_reducible_of_frobenii (M : Type*) (p : ℕ) [Fact (Nat.Prime p)] [AddCommGroup M]
+  [Module (ZMod p) M] [Module.Free (ZMod p) M] [Module.Finite (ZMod p) M]
+    (ρ : GaloisRep ℚ (ZMod p) M) (S : Finset (IsDedekindDomain.HeightOneSpectrum
+     (NumberField.RingOfIntegers ℚ))) (h : ∀ v ∉ S, 3 ∉ v.asIdeal → 5 ≤ (toNatPrime v).val →
+      ↑p ∉ v.asIdeal → LinearMap.charpoly ((ρ.toLocal v)
```
kbuzzard: Why put in all these extra hypotheses on v? You can just make S bigger to make them true. Why not just say that the char poly is (x-1)(x-v) for all v not in S?

I've merged `knownin1980s` so this would be a good use case. You can add the proof ("char poly is the same as char poly of 1+cyclo therefore same semisimplication by Brauer-Nesbitt") in a comment and then use `knownin1980s` instead of `sorry`.

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
```diff
@@ -13,13 +21,230 @@ noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p)
...
-    IsHardlyRamified P.hp_odd sorry
+    IsHardlyRamified P.hp_odd (EllipticCurve.torsion_has_rank2 P.freyCurve P.p (by sorry))
       (P.freyCurve.galoisRep P.p (show 0 < P.p from P.hppos)) :=
   sorry
 
 theorem FreyCurve.torsion_not_isIrreducible :
     haveI : Fact (P.p.Prime) := ⟨P.pp⟩
-    ¬ GaloisRep.IsIrreducible (P.freyCurve.galoisRep P.p P.hppos) :=
-  sorry -- TODO prove this
+    ¬ GaloisRep.IsIrreducible (P.freyCurve.galoisRep P.p P.hppos) := by
```
kbuzzard: Sorry, I have put the statement of this theorem in the wrong place :-( I now realise that we already have the statement in `FLT.Basic.Reductions`, which is I think where this argument should live.

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Well, there is a reason for that, but I realise in retrospect that that's not a good reason.
The function toNatPrime did not exist previously, it's something I had to define, so it has no API. Hence, I couldn't easily prove that the set {v : 3 ∉ v.asIdeal}, which is mathematically just {3}, is finite. I could work on developing the relevant API next, should it be in a separate file somewhere?

FLT/GaloisRepresentation/HardlyRamified/Frey.lean
stepan2698-cpu: Moved all of this reasoning into FLT.Basic.Reductions

stepan2698-cpu:
awaiting-review

FLT/Basic/Reductions.lean
```diff
@@ -35,19 +35,136 @@ open WeierstrassCurve
...
+/-- The natural `ℤ_p`-algebra structure on `ℤ/pℤ`. -/
+noncomputable local instance (p : ℕ) [Fact p.Prime] : Algebra ℤ_[p] (ZMod p) :=
+  RingHom.toAlgebra PadicInt.toZMod
+
+set_option linter.unusedVariables false in
+theorem irreducible_of_isCompatible_iff {K : Type*} [Field K] [NumberField K]
+  (E : Type*) [Field E] [NumberField E] (T : GaloisRepFamily K E 2) (comp : T.isCompatible)
+  {p : ℕ} [Fact p.Prime] {q : ℕ} [Fact q.Prime]
+  (φ : E →+* AlgebraicClosure ℚ_[p]) (ψ : E →+* AlgebraicClosure ℚ_[q]) :
+  GaloisRep.IsIrreducible (T _ φ) ↔ GaloisRep.IsIrreducible (T _ ψ) := knownin1980s
```
kbuzzard, line 60: I am still quite unclear about how you think this theorem is proved. Maybe you told me already but at least there should be a reference in the file.

Is what's going on here the following:
1) We want "hardly ramified mod 3 rep must be (cyclo *;0 1)" implies "hardly ramified semisimple 3-adic rep must be cyclo+1".
2) For some reason you don't like the proof in Daniel's notes
3) Using Ribet's Lemma instead you can show that a hardly ramified semisimple 3-adic rep must be reducible
4) We could now argue that these chars are flat at 3 and hence are unram and cyclo*unram at 3, and hence one is unram everywhere and thus trivial by Minkowski
5) Instead you want to invoke a theorem whose proof feels harder and for which you've given no reference.

My instinct is that the route via (4) is easier?

FLT/Basic/Reductions.lean
```diff
@@ -35,19 +35,136 @@ open WeierstrassCurve
...
+
+open GaloisRepresentation in
 theorem Wiles_Frey (P : FreyPackage) :
-    haveI : Fact P.p.Prime := ⟨P.pp⟩
-    ¬ GaloisRep.IsIrreducible (P.freyCurve.galoisRep P.p P.hppos) :=
-  FreyCurve.torsion_not_isIrreducible P
+  haveI : Fact P.p.Prime := ⟨P.pp⟩
```
kbuzzard: I made commit [42e3581](https://github.com/ImperialCollegeLondon/FLT/pull/761/commits/42e358162e1d1a73cd7a2d405e1b0ddbc387d0c2) to this PR which adds comments to the proof (all long proofs should be commented -- certainly anything this long) and changes various things to be more idiomatic. Unfortunately your indentation was wrong (your statement is indented 2 and your proof 4 instead of the other way around) so it's harder to see in the diff what changes I made. Other than the comments, the main things are `rw[` -> `rw [`, `have X := blah; rcases X with \<a,b,c,...\>` -> `obtain \<a,b,c,...\> := blah`, `rw [X], rw [Y]` -> `rw [X, Y]`, and don't name instances especially if you don't refer to them.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
 /-- Irreducibility of a Galois representation over a field. -/
 def GaloisRep.IsIrreducible {k : Type*} [Field k] [TopologicalSpace k] [Module k M]
     (ρ : GaloisRep K k M) : Prop := ρ.toRepresentation.IsIrreducible
+
+set_option linter.style.docString false in
```
kbuzzard: Fix the problem! Don't just silence the linter :-) Your spacing is bad.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
...
+/-- Note that we don't use  ρ.toRepresentation.IsAbsolutelyIrreducible because it will use
+the wrong type of base change. --/
```
kbuzzard: This is not an appropriate docstring. Every time someone sees `ρ.IsAbsolutelyIrreducible` and hovers over it to see what it means (e.g. did you go up to algebraic or separable closure, whatever) they will instead see this note and nothing else, rather than an explanation of what the declaration means. This should be a comment, not a doc comment. 

I am also unclear about what's going on here mathematically. Are you suggesting that for continuous representations there should be a different concept of absolute irreducibility? I had thought that for concepts like irreducibility and absolute irreducibility, the definition should just be "forget all topology and use the algebraic definition". That was my definition of `IsIrreducible`, for example.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
...
+    (ρ : GaloisRep K k M) : Prop where
+    absolutelyIrreducible : ∀ k', ∀ _ : Field.{u} k', ∀ _ : TopologicalSpace k',
+    ∀ _ : IsTopologicalRing k', ∀ _ : Algebra k k', ∀ _ : ContinuousSMul k k',
+      IsIrreducible (GaloisRep.baseChange k' ρ)
+
+namespace GaloisRep
+
+omit [NumberField K] [IsTopologicalRing A]
+lemma charpoly_baseChange [IsTopologicalRing B] [Algebra A B] [ContinuousSMul A B]
+ (ρ : GaloisRep K A M) (g : Field.absoluteGaloisGroup K) :
```
kbuzzard: Indent four spaces before the := and 2 after, in this and all following lemmas.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
...
+  TensorProduct R T V ≃ₗ[T] TensorProduct S T (TensorProduct R S V) := by exact
+    (TensorProduct.AlgebraTensorModule.cancelBaseChange R S T T V).symm
+
+theorem base_change_trans {K R : Type} (S T : Type*) [Field K] [TopologicalSpace R] [CommRing R]
+  [IsTopologicalRing R] [CommRing S] [TopologicalSpace S] [IsTopologicalRing S] [TopologicalSpace T]
+  [CommRing T] [IsTopologicalRing T] [Algebra R S] [ContinuousSMul R S] [Algebra S T] [Algebra R T]
+  [ContinuousSMul S T] [IsScalarTower R S T] [ContinuousSMul R T]
+  {V : Type*} [AddCommGroup V] [Module R V] [Module.Finite R V] [Module.Free R V]
+  (ρ : GaloisRep K R V) : GaloisRep.conj (GaloisRep.baseChange T ρ) (tensor_associator S T V) =
+  (GaloisRep.baseChange T (GaloisRep.baseChange S ρ)) := sorry
```
kbuzzard:
```suggestion
    (GaloisRep.baseChange T (GaloisRep.baseChange S ρ)) := by
  ext
  simp [tensor_associator]
```

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
...
+  [Module A M] [IsTopologicalRing A] [Module.Finite A M] [Module.Free A M]
+  (ρ : GaloisRep K A M) (g : Field.absoluteGaloisGroup K) :
+    (GaloisRep.baseChange B ρ).det g = Algebra.algebraMap (ρ.det g) := by
+  apply LinearMap.det_baseChange
+
+theorem irreducible_of_irreducible_base_change {K E F : Type*} [Field K] [Field E] [Field F]
+  [TopologicalSpace E] [IsTopologicalRing E] [TopologicalSpace F] [IsTopologicalRing F]
+  [Algebra E F] {V : Type*} [AddCommGroup V] [Module E V] [Module.Finite E V] [ContinuousSMul E F]
+  (ρ : GaloisRep K E V) : GaloisRep.IsIrreducible (GaloisRep.baseChange F ρ) →
+  GaloisRep.IsIrreducible ρ := sorry
```
kbuzzard, line 452: If we had base change for usual reps then this would follow from the claim that base change induced an injection from invariant submodules downstairs to invariant submodules upstairs.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
```diff
@@ -377,3 +380,72 @@ def GaloisRep.toRepresentation (ρ : GaloisRep K A M) : Representation A (Γ K)
...
+
+theorem reducible_conj_reducible_iff {K R : Type*} [Field K] [TopologicalSpace R] [Field R]
+  [IsTopologicalRing R] {V W : Type*} [AddCommGroup V] [Module R V] [Module.Finite R V]
+  [AddCommGroup W] [Module R W] [Module.Finite R W] [Module.Free R W]
+  (ρ : GaloisRep K R V) (e : V ≃ₗ[R] W) : GaloisRep.IsIrreducible ρ ↔
+  GaloisRep.IsIrreducible (GaloisRep.conj ρ e) := by sorry
```
kbuzzard, line 458: Again this follows from the analogous result for representations which is the generality in which this should be stated.

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
 
+theorem baseChange_hardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
+    {R : Type u} [CommRing R] [TopologicalSpace R] [IsTopologicalRing R] [IsLocalRing R]
+    [Algebra ℤ_[ℓ] R]
+    (S : Type u) [CommRing S] [TopologicalSpace S] [IsTopologicalRing S] [IsLocalRing S]
+    [Algebra R S] [Algebra ℤ_[ℓ] S] [ContinuousSMul R S] [IsScalarTower ℤ_[ℓ] R S]
+    {V : Type*} [AddCommGroup V] [Module R V]
+    [Module.Finite R V] [Module.Free R V] (hdim : Module.rank R V = 2)
+    (ρ : GaloisRep ℚ R V) : IsHardlyRamified hℓOdd hdim ρ →
+      IsHardlyRamified hℓOdd (by rw[Module.rank_baseChange, hdim]; exact Cardinal.lift_two)
```
kbuzzard:
```suggestion
      IsHardlyRamified hℓOdd (by rw [Module.rank_baseChange, hdim]; exact Cardinal.lift_two)
```

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+      (GaloisRep.baseChange S ρ) := sorry
```
kbuzzard, line 122: I am quite unclear about whether this is true. I am worried about flatness. How does the argument work here?

I am not sure I want a definition of hardly ramified which applies to representations to e.g. GL_2(Q_p).

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+    {W : Type*} [AddCommGroup W] [Module R W]
+    [Module.Finite R W] [Module.Free R W] (hdimW : Module.rank R W = 2)
+    (ρ : GaloisRep ℚ R V) (σ : GaloisRep ℚ R W)
+    (e : (FractionRing R) ⊗[R] V ≃ₗ[FractionRing R] (FractionRing R) ⊗[R] W)
+    (he : GaloisRep.conj (GaloisRep.baseChange (FractionRing R) ρ) e =
+      (GaloisRep.baseChange (FractionRing R) σ)) :
+    IsHardlyRamified hℓOdd hdimV ρ ↔ IsHardlyRamified hℓOdd hdimW σ := knownin1980s
+
+noncomputable example : ℚ →+* ℝ := by exact Rat.castHom ℝ
+
```
kbuzzard:
```suggestion
```

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+noncomputable def complexConjugationReal : Γ ℝ := sorry
+
+theorem complexConjugationReal_order_two : orderOf complexConjugationReal = 2 := sorry
+
+noncomputable def complexConjugation : Γ ℚ := (Field.absoluteGaloisGroup.mapAux (Rat.castHom ℝ))
+  complexConjugationReal
+
+theorem complexConjugation_order_two : orderOf complexConjugation = 2 := by
+  rw[orderOf_eq_prime_iff]
```
kbuzzard:
```suggestion
  rw [orderOf_eq_prime_iff]
```
etc etc (throughout the PR)

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+    [Module.Finite R W] [Module.Free R W] (hdimW : Module.rank R W = 2)
+    (ρ : GaloisRep ℚ R V) (σ : GaloisRep ℚ R W)
+    (e : (FractionRing R) ⊗[R] V ≃ₗ[FractionRing R] (FractionRing R) ⊗[R] W)
+    (he : GaloisRep.conj (GaloisRep.baseChange (FractionRing R) ρ) e =
+      (GaloisRep.baseChange (FractionRing R) σ)) :
+    IsHardlyRamified hℓOdd hdimV ρ ↔ IsHardlyRamified hℓOdd hdimW σ := knownin1980s
+
+noncomputable example : ℚ →+* ℝ := by exact Rat.castHom ℝ
+
+noncomputable def complexConjugationReal : Γ ℝ := sorry
```
kbuzzard, line 153: Probably the easiest thing to do is just to write down an isomorphism `AlgebraicClosure \R \equiv_a[\R] \C`

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+    rw[← map_pow, ← complexConjugationReal_order_two, pow_orderOf_eq_one, map_one]
+  · sorry
+
+theorem odd_of_hardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
+    {R : Type u} [CommRing R] [TopologicalSpace R] [IsTopologicalRing R] [IsLocalRing R]
+    [Algebra ℤ_[ℓ] R] {V : Type*} [AddCommGroup V] [Module R V]
+    [Module.Finite R V] [Module.Free R V] (hdim : Module.rank R V = 2)
+    (ρ : GaloisRep ℚ R V) (hρ : IsHardlyRamified hℓOdd hdim ρ) : GaloisRep.det ρ complexConjugation
+    = -1 :=
+      sorry
```
kbuzzard: I feel like the sorry here should be an assertion about the value of the cyclotomic character on complex conjugation rather than about hardly ramified Galois reps.

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
```diff
@@ -108,4 +110,72 @@ structure IsHardlyRamified {ℓ : ℕ} [Fact ℓ.Prime] (hℓOdd : Odd ℓ)
...
+
+theorem isAbsolutelyIrreducible_of_irreducible_odd {R : Type*} [TopologicalSpace R] [Field R]
+  [IsTopologicalRing R] {V : Type*} [AddCommGroup V] [Module R V] [Module.Finite R V]
```
kbuzzard: indent 4 for theorems before the := (everywhere in the PR)

FLT/GaloisRepresentation/HardlyRamified/ModThree.lean
```diff
@@ -6,15 +6,20 @@ local notation3 "Γ" K:max => Field.absoluteGaloisGroup K
 
 universe u
 
+def has_trivial_quotient (k : Type*) {K : Type*} [Field k] [Field K] {V : Type*}
```
kbuzzard, line 10: Give this a docstring

FLT/GaloisRepresentation/HardlyRamified/Threeadic.lean
```diff
@@ -1,24 +1,187 @@
...
+universe u
+
+open scoped TensorProduct
+
+/-- `toNatPrime v` returns the underlying rational prime of the height-one prime `v`
+of `ℤ = NumberField.RingOfIntegers ℚ`.
+-/
+noncomputable def toNatPrime
+  (v : IsDedekindDomain.HeightOneSpectrum (NumberField.RingOfIntegers ℚ)) :
+  {p : ℕ // Nat.Prime p} := by
```
kbuzzard: This is called `Nat.Primes`.

FLT/GaloisRepresentation/HardlyRamified/Threeadic.lean
```diff
@@ -1,24 +1,187 @@
...
 local notation "Frob" => Field.AbsoluteGaloisGroup.adicArithFrob
+local notation3 "Γ" K:max => Field.absoluteGaloisGroup K
 
 -- TODO -- make some API for "I have a rank 1 quotient where Galois acts trivially"
 -- e.g. this implies trace(Frob_p) is (1+p)
 
 /--
 A 3-adic hardly ramified representation has trace(Frob_p) = 1 + p for all p ≠ 2,3
 -/
+instance {R : Type*} [CommRing R] [TopologicalSpace R] [IsTopologicalRing R]
```
kbuzzard: I think that these are too contentious to be global instances. Make them local instances and then switch them in on files where you need them.

FLT/GaloisRepresentation/HardlyRamified/Threeadic.lean
```diff
@@ -1,24 +1,187 @@
...
+    [IsLocalRing R] [CompactSpace R] [T2Space R] [IsNoetherianRing R] :
+    DiscreteTopology (IsLocalRing.ResidueField R) := by
+    rw[discreteTopology_iff_isOpen_singleton_zero]
+    have := (IsModuleTopology.isOpenQuotientMap_of_surjective (B := IsLocalRing.ResidueField R)
+    (φ := {
+      toFun := (IsLocalRing.residue R).toFun
+      map_add' := (IsLocalRing.residue R).map_add'
+      map_smul' := by intro m x
+                      simp only [RingHom.toMonoidHom_eq_coe, OneHom.toFun_eq_coe,
```
kbuzzard:
```suggestion
      map_smul' := by
        intro m x
        simp only [RingHom.toMonoidHom_eq_coe, OneHom.toFun_eq_coe,
```
etc

FLT/Basic/Reductions.lean
stepan2698-cpu: The idea here is that I can skip step 4 for the following reason: If rho is a direct sum of two characters, then these characters must be cyclo^n \chi and cyclo^m \psi. Then we can run the exact same argument: charproly of Frob_l in the 3-adic is l^n \chi(l) + l^m \psi(l), hence charpoly of Frob_l in the p-adic is l^n \chi(l) + l^m \psi(l), hence p-adic is reducible. All of this was standard by the 1980s.
Now the reason why these characters must be this can be one of two things: we can either cite [Serre, "Abelian l-adic Representations and Elliptic Curves, III.3] for the spectacular result that any continuous p-adic character of G_Q is cyclo^n \chi, or we can say that because of the way this compatible family comes about (Tate modules of Shimura curves or something) our T is a compatible family of Hodge-Tate representations by construction.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
stepan2698-cpu: The problem with doing the obvious thing is that I would like to say "rho is absolutely irreducible, therefore rho tensor K is irreducible", and if I do it under the obvious definition, then Lean replies with "Representation.IsIrreducible (Representation.baseChange K rho.toRepresentation)", whereas I want  "GaloisRep.IsIreducible (GaloisRep.baseChange K rho)".
Mathematically I agree, as long as we are talking about finite-dimensional representations at least, absolute irreducibility should be equivalent to absolute irreducibility of rho.representation. Maybe I can prove this a theorem?

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
stepan2698-cpu: I'm not saying anything about Q_l. R and S here (and in the definition of IsHardlyRamified) are finite Z_l-algebras.

FLT/Deformations/RepresentationTheory/GaloisRep.lean
stepan2698-cpu: Fixed it

FLT/GaloisRepresentation/HardlyRamified/Defs.lean
stepan2698-cpu: Fixed it

FLT/Deformations/RepresentationTheory/GaloisRep.lean
stepan2698-cpu: Should I put the corresponding representation facts in FLT.Deformations.RepresentationTheory.Irreducible and deduce my facts from those facts? Or should I get rid of the GaloisRep facts altogether?

PR #773, opened 2025-11-27
`v.adicCompletion ℚ` is locally compact

smmercuri:


FLT/NumberField/Completion/Finite.lean
```diff
@@ -42,3 +44,7 @@ instance NumberField.instCompactSpaceAdicCompletionIntegers :
 lemma NumberField.isOpenAdicCompletionIntegers :
     IsOpen (v.adicCompletionIntegers K : Set (v.adicCompletion K)) :=
   Valued.isOpen_valuationSubring _
+
+instance (priority := 1) Rat.adicCompletion.locallyCompactSpace (v : HeightOneSpectrum (𝓞 ℚ)) :
```
kbuzzard: Why priority 1? (a) the repo compiles without this and (b) it seems like a great instance. `
#discr_tree_key Rat.adicCompletion.locallyCompactSpace` gives

```lean
LocallyCompactSpace
  (UniformSpace.Completion
     (@WithVal ℚ (WithZero (Multiplicative ℤ)) _ _ (@HeightOneSpectrum.valuation (𝓞 ℚ _) _ _ ℚ _ _ _ _))
     _)
  _
```
which I think says something like "this instance will never be applied unless the goal looks something like this" (so it will hardly ever be applied) and conversely if the goal looks something like that then you probably want to try this instance first rather than last.

FLT/QuaternionAlgebra/NumberField.lean
```diff
@@ -258,6 +258,13 @@ theorem GL2.TameLevel.isOpen : IsOpen (GL2.TameLevel S).carrier :=
 theorem GL2.TameLevel.isCompact : IsCompact (GL2.TameLevel S).carrier :=
   sorry
 
+-- introduction of `LocallyCompactSpace (v.adicCompletion ℚ)` instance in
+-- `FLT.NumberField.Completion.Finite`causing timeouts in instance search for
+-- `IsScalarTower F (FiniteAdeleRing (𝓞 F) F) (FiniteAdeleRing (𝓞 F) F)`
+-- and `ContinuousSMul (FiniteAdeleRing (𝓞 F) F) (Matrix _ _ (FiniteAdeleRing (𝓞 F) F)`
+attribute [local instance high] IsScalarTower.right
+attribute [local instance high] IsModuleTopology.toContinuousSMul
```
kbuzzard: I think that the correct fix isn't more hacks, it's to make `FiniteAdeleRing` a def and not an abbrev. I start on this in mathlib #32277 . Unfortunately this breaks FLT (prior to this PR): attempted fix is at #784

FLT/NumberField/Completion/Finite.lean
smmercuri: Oh woops, I forgot to revert that. Was testing some things to try to resolve the other timeouts.

FLT/QuaternionAlgebra/NumberField.lean
smmercuri: Yes, that makes sense as a fix. I also have a WIP PR from a while ago doing the same to `adicCompletion` for similar reasons

kbuzzard: Ok FiniteAdeleRing is a def on `main` now (and in mathlib master), maybe merge `main` and see if it reduces the number of hacks you need?

smmercuri: Unfortunately, the hacks were still required. After digging into the instance search traces, turns out the culprit to the slowdown was the transitive import of `FLT.DedekindDomain.FiniteAdeleRing.BaseChange`, and in particular [this `IsScalarTower` instance](https://github.com/ImperialCollegeLondon/FLT/blob/3365f27d067c4b02c684ea5d6a201a1c44abca00/FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean#L103-L108) causing issues with `A = B; K = L`. 

Anyway, hacks have been removed after tidying up imports.

kbuzzard: I see. Thanks a lot for the diagnosis! This diamond comes from `Algebra K L -> Algebra (\A K) (\A L)` where \A is adeles or finite adeles or infinite adeles etc; if K=L this is problematic, but I don't see how to get around it.

PR #783, opened 2025-12-01
feat: add proof of E_noninjective_right

kbuzzard:
Also add some \leanok's in the blueprint which I noticed when working on this.

Also rearrange some of the results in the file so we can avoid `omit`.

Also min_imports.


FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -144,6 +149,7 @@ abbrev incl_Kn_𝔸Kn : (Fin (Module.finrank K D) → K) →
     (Fin (Module.finrank K D) → AdeleRing (𝓞 K) K) :=
   fun x i ↦ algebraMap K (AdeleRing (𝓞 K) K) (x i)
 
+-- these must be in the wrong place?
```
kbuzzard: I tried to move them but got timeouts. Maybe wait until the FiniteAdeleRing refactor #784 lands.

PR #796, opened 2025-12-08
Fujisaki : Units_cocompact

WilliamCoram:
We prove the sorry in `NumberField.FiniteAdeleRing.DivisionAlgebra.units_cocompact`.

### DIFF
--- a/FLT/DivisionAlgebra/Finiteness.lean
+++ b/FLT/DivisionAlgebra/Finiteness.lean
@@ -418,6 +418,7 @@ lemma ImAux_isCompact : IsCompact ((fun p ↦ (p.1, MulOpposite.op p.2)) '' Aux.
 lemma M_compact : IsCompact (M K D) := Topology.IsClosedEmbedding.isCompact_preimage
   (incl₂_isClosedEmbedding K D) (ImAux_isCompact K D)
 
+-- Thomas needs this IIRC
 lemma compact_quotient [Algebra.IsCentral K D] :
     CompactSpace (_root_.Quotient (QuotientGroup.rightRel
     ((MonoidHom.range (incl K D)).comap (ringHaarChar_ker D_𝔸).subtype))) :=
@@ -444,7 +445,7 @@ abbrev Dinf := D ⊗[K] (NumberField.InfiniteAdeleRing K)
 abbrev Dinfx := (Dinf K D)ˣ
 
 -- this instance creates a nasty diamond for
--- `IsScalarTower K (FiniteAdeleRing A K) (FiniteAdeleRing B L)` when K = L A = B, and
+-- `IsScalarTower K (FiniteAdeleRing A K) (FiniteAdeleRing B L)` when K = L and A = B, and
 -- should probably be scoped (or even removed and statements changed so that they
 -- don't need it).
 attribute [-instance] instIsScalarTowerFiniteAdeleRing_fLT_1
@@ -551,8 +552,8 @@ local instance : Algebra ℝ (Dinf K D) :=
     intro c x
     rw [RingHom.comp_apply, Algebra.commutes]
 
-local instance : IsScalarTower ℝ (InfiniteAdeleRing K) (Dinf K D) := by
-  exact IsScalarTower.of_algebraMap_eq (congrFun rfl)
+local instance : IsScalarTower ℝ (InfiniteAdeleRing K) (Dinf K D) :=
+  IsScalarTower.of_algebraMap_eq (fun _ ↦ rfl)
 
 local instance : Module.Finite ℝ (Dinf K D) :=
   Module.Finite.trans (InfiniteAdeleRing K) (Dinf K D)
@@ -579,6 +580,7 @@ local instance : MeasurableSpace (Df K D) := borel (Df K D)
 
 local instance : BorelSpace (Df K D) := { measurable_eq := rfl }
 
+-- D ⊗ K_∞ is second countable because it's a finite ℝ-module
 instance : SecondCountableTopology (Dinf K D) :=
   Module.Finite.secondCountabletopology ℝ (Dinf K D)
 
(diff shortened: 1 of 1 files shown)


FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -348,10 +366,249 @@ open scoped TensorProduct.RightActions
...
+local instance : MeasurableSpace (Dinf K D) :=
+  borel (Dinf K D)
+
+local instance : BorelSpace (Dinf K D) := {measurable_eq := rfl }
+
+local instance : MeasurableSpace (Df K D) := borel (Df K D)
+
+local instance : BorelSpace (Df K D) := { measurable_eq := rfl }
+
+local instance : MeasurableSpace (Dinf K D × Df K D) := borel (Dinf K D × Df K D)
```
kbuzzard: This creates a diamond; if X and Y have measurable space structures then X x Y gets one from something like `Prod.instMeasurableSpace`, so you can't have this instance. It should be there anyway, and it will also satisfy BorelSpace automatically as long as you've proved the factors are second countable.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -323,19 +323,37 @@ section FiniteAdeleRing
...
+
 -- Instance to help speed up instance synthesis
-instance : NonUnitalNonAssocRing (D ⊗[K] (FiniteAdeleRing (𝓞 K) K)) :=
+instance : NonUnitalNonAssocRing (Df K D) :=
   let r := Algebra.TensorProduct.instRing.toNonUnitalRing
   r.toNonUnitalNonAssocRing
 
 -- Instance to help speed up instance synthesis
-instance : NonAssocSemiring (D ⊗[K] (FiniteAdeleRing (𝓞 K) K)) :=
+instance : NonAssocSemiring (Dinf K D) :=
```
kbuzzard: Is this really needed? I made FiniteAdeleRing irreducible in mathlib and in this repo to solve problems like this. Are you up to date with `main` in this branch and if not then maybe try merging `main` and seeing if it solves your problems? I am quite averse to merging hacks and a bunch of things here feel like hacks.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -348,10 +366,249 @@ open scoped TensorProduct.RightActions
...
 
+/-- Notation for (Algebra.TensorProduct.prodRight K K D (NumberField.InfiniteAdeleRing K)
+    (FiniteAdeleRing (𝓞 K) K)). -/
+abbrev D𝔸_prodRight : D_𝔸 ≃ₐ[K] Dinf K D × Df K D :=
+  (Algebra.TensorProduct.prodRight K K D (NumberField.InfiniteAdeleRing K)
+    (FiniteAdeleRing (𝓞 K) K))
+
+local instance : Module (AdeleRing (𝓞 K) K)
+    (D ⊗[K] (InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K)) :=
+  TensorProduct.RightActions.instModule_fLT K (InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K) D
```
kbuzzard: AdeleRing (𝓞 K) K is *defined* to be (D ⊗[K] (InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K)). Are you sure you need to break apart this definition? The whole point of defining `AdeleRing` is so that you never have to deal with `InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K`.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -348,10 +366,249 @@ open scoped TensorProduct.RightActions
...
+
+/-- The restriction of ringHaarChar_ker D_𝔸 to (D ⊗ 𝔸_K^∞)ˣ via D𝔸_iso_prod_units. -/
+abbrev rest₁ : ringHaarChar_ker D_𝔸 → Dfx K D :=
+  fun a => (D𝔸_prodRight_units K D) a.val |>.2
+
+omit [Algebra.IsCentral K D] in
+lemma rest₁_continuous : Continuous (rest₁ K D) := Continuous.comp continuous_snd (Continuous.comp
+  (D𝔸_prodRight_units_cont K D) continuous_subtype_val)
+
+local instance : Algebra ℝ (Dinf K D) := by
```
kbuzzard: This needs to be filled in as it's data, so filling it in might break proofs.

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: I am up to date (as of this morning) on main - all the instance's included are because at some point when writing a statement Lean was complaining they could not find an instance for it - if you delete any one you will see where it breaks.

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: I remember not being able to find anything - but perhaps I did not look hard enough. I will ask around on Zulip to see if this exists.

FLT/DivisionAlgebra/Finiteness.lean
WilliamCoram: Although I mentioned earlier that all instances were required - apparently this can be removed. I still need the instance below though so that we can get the module structure on the product, perhaps I could format it in one instance by using .1 and .2 of elements of AdeleRing?

WilliamCoram: I have adjusted the module instance to remove all `InfiniteAdeleRing K × FiniteAdeleRing (𝓞 K) K` appearances.

FLT/DivisionAlgebra/Finiteness.lean
```diff
@@ -392,10 +411,306 @@ open scoped TensorProduct.RightActions
...
+/-- Dinf K D has the ℝ-module topology. -/
+local instance : IsModuleTopology ℝ (Dinf K D) := by
+  /- By Algebra ℝ (InfiniteAdeleRing K); (InfiniteAdeleRing K) has the ℝ-module topology.
+    Now since (Dinf K D) has the (InfiniteAdeleRing K)-module topolology it also has the
+    ℝ-module topology.
+  -/
+  have : IsModuleTopology ℝ (InfiniteAdeleRing K) := by
+    sorry
+  --have : IsModuleTopology (InfiniteAdeleRing K) (Dfx K D) := by
+  -- this is not being found
```
kbuzzard: This doesn't even make sense because `Dfx K D` isn't a module for `InfiniteAdeleRing K` (it's not even an additive group)

WilliamCoram: This should read Dinf K D - thats a typo on my end. You're right that this obviously doesnt make sense as written

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: Now fixed after locally removing bad instance

FLT/DivisionAlgebra/Finiteness.lean
kbuzzard: We have this in the PR now.

PR #799, opened 2025-12-13
feat: prove `one_smul` for galoisRepresentation

MrBrain295:
I'm sorry if I did something wrong. I wanted to see if AI could prove tiny parts and it can (I think)! I used Google Gemini and after telling it the errors lean gave me a few times it came up with this and lean accepted it.

### DIFF
--- a/FLT/EllipticCurve/Torsion.lean
+++ b/FLT/EllipticCurve/Torsion.lean
@@ -102,8 +102,11 @@ noncomputable instance WeierstrassCurve.galoisRepresentation_smul
 noncomputable def WeierstrassCurve.galoisRepresentation
     (K : Type u) [Field K] [DecidableEq K] [Algebra k K] :
     DistribMulAction (K ≃ₐ[k] K) (E ⟮K⟯) where
-      one_smul := sorry -- these should all be easy
-      mul_smul := sorry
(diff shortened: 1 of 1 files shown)


FLT/EllipticCurve/Torsion.lean
```diff
@@ -102,8 +102,11 @@ noncomputable instance WeierstrassCurve.galoisRepresentation_smul
 noncomputable def WeierstrassCurve.galoisRepresentation
     (K : Type u) [Field K] [DecidableEq K] [Algebra k K] :
     DistribMulAction (K ≃ₐ[k] K) (E ⟮K⟯) where
-      one_smul := sorry -- these should all be easy
-      mul_smul := sorry
+      one_smul := fun P ↦ by
+        change WeierstrassCurve.Points.map E (AlgHom.id k K) P = P
+        rw [WeierstrassCurve.Points.map_id E K]
+        rfl
```
kbuzzard, line 108: Yes it works, but what really needs to be done here is that we need to change `WeierstrassCurve.Points.map` to `WeierstrassCurve.Affine.Point.map` now mathlib has it (it didn't have it when I wrote this code originally) and adapt to the new mathlib spelling of everything, rather than going ahead building more stuff which will need to be refactored.

PR #804, opened 2025-12-18
refactor: rewrite Frobenius project chapter

tb65536:
This PR rewrites the Frobenius project chapter, basically following the structure of `Mathlib/RingTheory/Invariant/Basic.lean`.

### DIFF
--- a/blueprint/src/chapter/FrobeniusProject.tex
+++ b/blueprint/src/chapter/FrobeniusProject.tex
@@ -188,8 +188,8 @@ it's not even well-defined if the map $A\to B$ isn't injective.
 \end{proof}
 
 \begin{theorem}
-  \label{MulSemiringAction.CharacteristicPolynomial.isIntegral}
-  \lean{MulSemiringAction.CharacteristicPolynomial.isIntegral}
+  \label{Algebra.IsInvariant.isIntegral}
+  \lean{Algebra.IsInvariant.isIntegral}
   \mathlibok
   $B/A$ is integral.
 \end{theorem}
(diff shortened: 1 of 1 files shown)


tb65536:
The theorem labels have been changed to mathlib declaration names, but one thing I'm not sure how to do (or if this is even feasible/desirable) is actually linking this up with mathlib.

tb65536:
awaiting-review

kbuzzard:
Just to be clear on this -- my vision for the FLT blueprint is that it is going to be a top-down exposition of FLT and I am going to be deleting chapters 8 to 11 once they are fully green, like I deleted this one (it's not being rendered on the web version). I don't really know what to do with this stuff; my main problem is that I would love to put these chapters away in a place by themselves so that they don't interfere with what I want to be the "main" FLT graph (which I now have a vision for) but right now the blueprint is limited to only being able to produce one big blueprint (which will be too big/disconnected unless I delete a ton of chapters) or one blueprint per chapter (which we have right now and then it's difficult to do cross-chapter links meaningfully).

blueprint/src/chapter/FrobeniusProject.tex
```diff
@@ -177,314 +172,168 @@ \section{The extension \texorpdfstring{$B/A$}{B/A}.}
...
-  $\overline{M}_{\overline{b}}$ has degree $|G|$.
-\end{theorem}
+\begin{lemma}
+  \label{fixed_of_fixed1_aux1}
+  \mathlibok
+  There exist elements $a,b \in B$, with $a$ in the image of $A$, such that for all $g\in G$,
+  \begin{itemize}
+    \item If $g \cdot Q = Q$, then $g \cdot b \equiv a \pmod{Q}$.
+    \item If $g \cdot Q \neq Q$, then $g \cdot b \equiv 0 \pmod{Q}$.
+  \end{itemize}
```
kbuzzard, line 217: Better proof: a=b=0. Same for next lemma. Have I misunderstood something?

tb65536: Ah, I missed a hypothesis `a ∉ Q`.

tb65536:
awaiting-review

PR #805, opened 2025-12-18
feat: ringHaarChar_adeles_units_rat_eq_one

bwangpj:


kbuzzard:
Nice! This looks ready to go. Are you using continuous Z-algebra equivs because we don't have continuous ring equivs? If so, can you add a comment pointing this out?

bwangpj:
> Nice! This looks ready to go. Are you using continuous Z-algebra equivs because we don't have continuous ring equivs? If so, can you add a comment pointing this out?

Yep, and added

PR #816, opened 2025-12-24
`padicEquiv` is an isometry

smmercuri:
- Proof of `padicEquiv_norm_eq` and move to `Mathlib` folder
- Move `intValuation_eq_coe_neg_multiplicity` to `Mathlib` folder to preserve `FLT.Mathlib` imports


smmercuri:
Not sure if I'm doing the right thing to bump mathlib here. Added `rev = <hash>` to the lakefile to get it to work locally, but since reverted that change and only kept `lake-manifest.json` changes in the PR

kbuzzard:
(I bumped mathlib on main)

PR #877, opened 2026-04-01
chore: bump mathlib to lean-4.29.0

kbuzzard:
It's time to bump now there's a new stable version of Lean.


FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean
```diff
@@ -15,7 +15,7 @@ import Mathlib.RingTheory.Flat.TorsionFree
 
 # Base change of adele rings.
 
-If `A` is a Dedekind domain with field of fractions `K`, if `L/K` is a finite separable
+If `A` is a Dedekind domain with field of fractions `K`, if `L/Ktha` is a finite separable
```
kbuzzard:
```suggestion
If `A` is a Dedekind domain with field of fractions `K`, if `L/K` is a finite separable
```

kbuzzard:
CI is green and I'm doing one last review. I propose merging all the technical debt so we can get on (we should ideally bump again). There are three main issues:

1) occasional `backward.isDefEq.respectTransparency false` throughout which we'll have to come back to;
2) huge slowdowns in several files (which I understand the cause of but cannot fix -- see issue https://github.com/ImperialCollegeLondon/FLT/issues/889 );
3) `backward.isDefEq.respectTransparency false` essentially everywhere in `FLT/DivisionAlgebra/Finiteness.lean` which I've just asked about on Zulip [#general > backward.isDefEq.respectTransparency @ 💬](https://leanprover.zulipchat.com/#narrow/channel/113488-general/topic/backward.2EisDefEq.2ErespectTransparency/near/585655500) (solution: this is caused by a bad instance in mathlib)

PR #882, opened 2026-04-06
fix: `WithAbs` issues in 4.29.0 bump

smmercuri:
- `WeakApproximation.lean` is now in mathlib 
- `Dimension.lean` is almost in mathlib, so just replace contents with the open mathlib PR [#30551](https://github.com/leanprover-community/mathlib4/pull/30551)
- most of `WithAbs.lean` is in mathlib
- Introduce `SemialgHom.restrictScalars` to go from `v.Completion →ₛₐ[algebraMap (WithAbs v.1) (WithAbs w.1)] w.Completion` to `v.Completion →ₛₐ[algebraMap K L] w.Completion` which was previously defeq


kbuzzard:
Thanks!

PR #887, opened 2026-04-13
fix: finiteness files in `4.29` bump

smmercuri:
- A couple more `maxHeartbeats 0` required
- Remove upstreamed material in `MeasureTheory.Haar.Extension`


kbuzzard:
Wow, we must have committed some kind of terrible sin in that finiteness file!

PR #895, opened 2026-04-15
chore(FLT/Mathlib/MeasureTheory/Haar/Extension): clean up file post-bump

tb65536:
There were some extraneous variables lines and theorems after the bump.


kbuzzard:
Many thanks!

PR #898, opened 2026-04-16
feat(Patching/Utils/CompactHausdorffRings): add partial proof that compact Hausdorff rings are totally disconnected

tb65536:
I don't have immediately plans to finish this off (since @YaelDillies is actively working in this area and has ideas for how to do things in the right generality for mathlib), so I'm putting the current state of things here.

### DIFF
--- a/FLT.lean
+++ b/FLT.lean
@@ -161,6 +161,7 @@ import FLT.Patching.REqualsT
 import FLT.Patching.System
 import FLT.Patching.Ultraproduct
 import FLT.Patching.Utils.AdicTopology
+import FLT.Patching.Utils.CompactHausdorffRings
 import FLT.Patching.Utils.Depth
 import FLT.Patching.Utils.InverseLimit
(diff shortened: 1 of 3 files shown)


YaelDillies:
Can you record somewhere in the code that this is being worked on over at YaelDillies/mean-fourier ?

tb65536:
> Can you record somewhere in the code that this is being worked on over at YaelDillies/mean-fourier ?

Thanks! I've added this.

kbuzzard:
Thanks! Do we have a way of saying "I am a projective limit of finite rings"? (I'm happy to merge as it is but this seemed to be what I was using in the FLT course, rather than "I am a projective limit of finite sets").

tb65536:
Not that I've seen, although for local rings FLT/Patching/Utils/AdicTopology provides some of the glue (e.g., powers of the maximal ideal form a basis for the topology).

kbuzzard:
Oh thanks! I only care about the local ring case I guess.

PR #910, opened 2026-04-26
fix(CI): Blueprint test driver

YanYablonovskiy:
Checking if the band aid works in the main repo, consists of a CI step to forcibly relocate the `FLTTest.olean` for the blueprint checkdecls to see.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -76,7 +76,6 @@ jobs:
         with:
           fetch-depth: 0 # Fetch all history for all branches and tags
 
-
       - name: Build and lint project
         uses: leanprover/lean-action@v1
         with:
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -88,7 +88,7 @@ jobs:
       - name: Adjusting for test copy
         if: always()
         run: |
-           cd .lake/build/lib/lean/FLTTest | copy "FLTTest*" .lake/build/lib/lean /Y
+           cp -f .lake/build/lib/lean/FLTTest/FLTTest* .lake/build/lib/lean/
 
       - name: Upstreaming dashboard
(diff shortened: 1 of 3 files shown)


kbuzzard:
I tried undrafting it to see if I could get CI running but I might just have to merge it?

.github/workflows/blueprint.yml
```diff
@@ -76,6 +76,7 @@ jobs:
         with:
           fetch-depth: 0 # Fetch all history for all branches and tags
 
+
```
kbuzzard: ?

FLTTest/FLTTest.lean
```diff
@@ -0,0 +1 @@
+import FLTTest.MathlibCompatibility
```
kbuzzard, line 1: We're now importing this twice?

kbuzzard:
Something I am slightly concerned about it this. The way the docs work is that the only mathlib docs which are created in FLT are the parts of mathlib which are imported. I definitely don't want to compile docs for all of mathlib in FLT CI because this will double the time it takes for CI to run on a PR to probably well over an hour.

FLTTest/FLTTest.lean
YanYablonovskiy: Unfortunately, I tried to change it but could not. We just need it to make an `FLTTest.olean` with the same imports (and content but there is none luckily) to match the one in the main directory.  

However if you add the file "FLTTest.lean" in the working directory (`FLTTest\FLTTest.lean`), but don't import `FLTTest.FLTTest` , it will complain about `mk_all`, which will then make you import it again. 

This is a bit of a redundant transitory import , essentially just a hack to get the right `.olean` and associated (which essentially just contains the imports, so `FLTTest.lean` from `\FLTTest\` has an agreeable `.olean` with the main dir `FLTTest.lean` governed by `mk_all`.

PR #926, opened 2026-05-03
don't allow noisy PRs to pass CI

### LINKED ISSUE #768: don't allow noisy PRs to pass CI
A couple of times now I have merged PRs with stuff like `#check` or `#lint` accidentally included. We should be linting for this like mathlib does.

Milian0402:
Closes #768.

This adds a small CI check for noisy interactive `#` commands such as `#check`, `#eval`, `#lint`, and `#print`, which was the issue raised in #768.

The check ignores comments and strings, allows `#adaptation_note`, and preserves the existing intentional `#print axioms` line.


SnirBroshi:
Since the CI already builds the code I don't think there's a reason to use a fallible Python script to grep for these.
`lake build` can do it for you, which is what Mathlib uses.
Though we'd need to add a `#guard_msgs` to the `#print axioms`.

kbuzzard:
After some thought I've decided to go with #975 even though it makes me a little sad -- I'd just rather have a community tool than a random python script which I don't really understand and will have to maintain.

Milian0402:
Fair enough, #975 is a cleaner way to do this! Thanks for sorting it out

PR #935, opened 2026-05-05
run CI

kbuzzard:
Empty commit to trigger CI.


kbuzzard:
(NB this doesn't work, so I'll make another PR actually changing a lean file)

PR #948, opened 2026-05-07
use patched version of docgen-action to prevent cache invalidation

dwrensha:
This should bring CI times back into a more reasonable range.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -98,7 +98,11 @@ jobs:
           include-drafts: true
 
       - name: Compile blueprint and documentation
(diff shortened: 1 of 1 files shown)


dwrensha:
The workaround includes these two commits: 
https://github.com/dwrensha/docgen-action/commit/4b04586b24d6b21ed0561df8fc356867dddccfe4
https://github.com/dwrensha/docgen-action/commit/a3f8189646977363717917b605e6de25f9704656

dwrensha:
Closing, because the `v4.30.0-rc2` tag of doc-gen4 no longer has this problem. See https://leanprover.zulipchat.com/#narrow/channel/536994-ecosystem-infrastructure/topic/huge.20increase.20in.20FLT.20doc-gen.20compile.20time/near/593566038.

PR #949, opened 2026-05-07
chore: modularize index files

SnirBroshi:
This is the result of `lake exe mk_all --module`, and thanks to leanprover-community/mathlib4#38395 `lake exe mk_all --check` in CI should still succeed, and further `mk_all`s even without `--module` will use the `module` format.
(The current Mathlib version already includes that PR.)


kbuzzard:
Thanks a lot! Let's give it a try!

PR #964, opened 2026-05-17
chore: bump mathlib to get Gouezel's def of Ring

kbuzzard:


kbuzzard:
Got trumped by the machine

PR #975, opened 2026-05-23
chore: fail CI on any output

### LINKED ISSUE #768: don't allow noisy PRs to pass CI

SnirBroshi:
Fails the build in CI on any output, including warnings (e.g. if a line exceeds 100 chars) and info messages (e.g. a stray `#check` or `#lint`).

This works by passing the flag [`--iofail`](https://github.com/leanprover/lean4/blob/ad1c983a43face43c3e69b72f7db105f40ef280a/src/lake/Lake/CLI/Help.lean#L72-L73) to `lake build`, like [Mathlib](https://github.com/leanprover-community/mathlib4/blob/50622aa047ff449311db5bbdaf31784bd95a1232/.github/workflows/build_template.yml#L489) and [Cslib](https://github.com/leanprover/cslib/blob/6cc5327b2a45ea7a317f1a4951f31c345a267c75/.github/workflows/lean_action_ci.yml#L23-L24) do.

Also adds a `#guard_msgs in` to the `#print axioms` in `FermatsLastTheorem.lean`, which otherwise would fail CI.

Fixes #768, see the [failed run](https://github.com/ImperialCollegeLondon/FLT/actions/runs/26330788639/job/77516438871?pr=975) after adding a `#lint`.

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -83,6 +83,8 @@ jobs:
           test: true
           lint: true
           mk_all-check: true
+          build-args: --iofail
+          test-args: --iofail
 
       - name: Adjusting for test copy
         if: always()
--- a/FLT/GaloisRepresentation/Automorphic.lean
+++ b/FLT/GaloisRepresentation/Automorphic.lean
@@ -180,5 +180,3 @@ theorem cyclic_base_change
     ((ρ.map (algebraMap F E)).IsAutomorphicOfLevel p hV
       (HeightOneSpectrum.preimageComapFinset (𝓞 F) F E (𝓞 E) S)) :=
   sorry
-
-#lint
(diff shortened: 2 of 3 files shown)


kbuzzard:
I like this a lot more than #926 but it makes me a little bit sad that I have to lose the reminder that I haven't proved FLT yet every time I compile the repo on the command line :-) Several times now I have tried to do something a little bit different in FLT and then there's been a PR which "fixes" things by making the repo follow the community norms, with the argument typically being that it's easier to apply community tools if I am following community norms. This PR is a case in point: I want to check for spurious `#lint` and `#check` and also deprecation warnings, but I like my "you haven't proved FLT" output; however the community tool says "well mathlib wants 0 output so here's a great tool which can check that for you too, sorry about the FLT joke but that has to go" :-(

kbuzzard:
OK having mulled it over, I think that probably this is the correct approach. Thanks Snir.

PR #977, opened 2026-05-25
fix: restore `#print axioms`

SnirBroshi:
Restore `#print axioms` (after #975) and instead delete the `#print axioms` line from `./FermatsLastTheorem.lean` before building in CI, so locally `lake build` will still print the axioms.


kbuzzard:
I think that probably I should "conform" and not do funky stuff like this :-(

PR #999, opened 2026-06-03
feat: add basic verso blueprint infrastructure

kbuzzard:
Add a test verso blueprint file.


kbuzzard:
IIRC there were some artificial mentions of nodes so I could draw arrows; I should be able to fix this now.

TODO:

[x] get this compiling again (it's currently broken)
[x] fix unfolding
[] fix the artificial arrow issues -- postponed

PR #1010, opened 2026-06-08
feat: add verso blueprint level 1

kbuzzard:
This time without bumping mathlib beyond v4.30.0 so we don't break versoblueprint...


kbuzzard:
Closed in favour of #1011

PR #1020, opened 2026-06-12
feat: classification of finite subgroups of PGL_2

kbuzzard:
By Duxing Yang, bumped from v4.26 to v4.31 by Claude.

PR #1029, opened 2026-06-19
feat: Tame case of Dickson classification

Yangdx02:
Tame case for Dickson classification done in PGL, CyclicPartition and TameDickson. Files PSL and RecognitionA5 are for the wild case.


kbuzzard:
Hi. Are you Duxing Yang? In which case you might be interested in #1020

kbuzzard:
General case now merged!

PR #1067, opened 2026-07-06
feat(Representation theory): Clifford theory statements

mckoen:
slop of three statements

- if rho is a 2-dimensional rep of G over an alg closed field and H is a finite index normal subgroup with cyclic quotient and such that rho|H is the direct sum of two equal chars then rho is reducible.

- if rho is a 2-dimensional irred rep of G over an alg closed field and if H is a finite index normal subgroup with G/H cyclic then rho|H is either irreducible or the direct sum of two G-conjugate characters.

- If rho is a 2-dimensional rep of G over an alg closed field and H is a finite index normal subgroup with cyclic quotient of even order and such that rho|H is the direct sum of two G-conjugate characters, then rho is Ind_K^G(psi) where K is the index 2 subgroup of G containing H and psi is a 1-d rep of K.


kbuzzard:
Is this still WIP?

kbuzzard:
Is this still WIP?

I bumped mathlib and it broke some of the code here. I fixed it in this commit https://github.com/ImperialCollegeLondon/FLT/commit/e8dc821f390dd2e5facc4fbf912b739b44733fa0 (not on your branch yet).

Deicyde:
The `clifford.lean` file relies on a definition in the slop folder:

- `SplitsAsDistinctConjugateCharacters` 

Please move these into a file `KnownIn1980s/def.lean` so that your theorem statement can be stated without dependencies on slop--or move the theorem `isIrreducible_comp_subtype_or_splitsAsDistinctConjugateCharacters` into slop.

Issue #1061, opened 2026-07-06
The Tate curve construction


kbuzzard:
Explicit power series X and Y satisfy a plane cubic equation. This fundamental equality, whose standard proof is "prove the relationship between the Weierstrass P-function and its derivative, and then take power series expansions", has been stated in FLT/KnownIn1980s/EllipticCurves/TateCurveConstruction.lean .

Note that a test to give evidence that there are no typos in the theorem statement is carried out in FLT/KnownIn1980s/EllipticCurves/TateCurveConstructionExperiment.lean , where the identity is checked to O(q^3).

PR #1069, opened 2026-07-06
feat: Construction of the Tate curve (#1061)

WilliamCoram:
This PR addresses #1061.

Co-authered by Samuel Yin


WilliamCoram:
In the spirit of burning credits I gave Fable the following prompt: 
"Can you golf and clean all the proofs in this file. If it looks like a have statement is general and can be extracted as its own lemma extract. Specifcally, try and make all proofs less than 50 lines"

Just this managed to cut the line count by ~500, and I am sure more would be possible with more robust prompts or Claude skills.

PR #1071, opened 2026-07-06
feat: Statement of Moret-Bailly' theorem

justus-springer:


FLT/KnownIn1980s/MoretBailly/RationalPoints.lean
```diff
@@ -0,0 +1,89 @@
+/-
+Copyright (c) 2026 Kevin Buzzard. All rights reserved.
```
kbuzzard, line 2: ?!

PR #1077, opened 2026-07-07
refactor: restate restrictNormal lemma using restrictNormalHom, golf proofs

pepamontero:
This PR makes some minor improvements to `FLT/Mathlib/FieldTheory/Galois/Basic.lean`:

- Renames `forall_apply_algebraMap_iff_restrictNormal_eq_one` to `forall_apply_algebraMap_iff_restrictNormalHom_eq_one`, restating it in terms of `AlgEquiv.restrictNormalHom` instead of `AlgEquiv.restrictNormal`.
  - This is the form `map_mul'` and `quadraticCharacter_surjective` actually need, so both proofs can use it directly instead of unfolding first.
  - Updates the lemma's one other call site, in `QuadraticTwists.lean` (which works by just changing to the lemma's name).
- Golfs some proofs in the file.

Note: the original version of this PR closed some sorrys of this file, but those were closed already so this contains small follow-up improvements on top of the merged code.


kbuzzard:
The plan here is to wait for Michael Stoll's more substantial PR and then we'll compare and contrast :-)

pepamontero:
Since #1088 has been merged, I've repurposed this PR into a small follow-up on top of it: it restates `forall_apply_algebraMap_iff_restrictNormal_eq_one` in terms of `AlgEquiv.restrictNormalHom` (renaming it to `forall_apply_algebraMap_iff_restrictNormalHom_eq_one`) and golfs a few proofs using it. Happy to adjust anything!

PR #1091, opened 2026-07-08
chore: golf TateCurveConstruction.lean

pepamontero:
Proof golfing pass over TateCurveConstruction.lean, no mathematical changes. Roughly -200 LoC.
Systematic changes:

- Replaced `have h := ...`; `rw [...] at h`; `exact h` patterns with `simpa [...] using h`
- Dropped `show ... from by ...` in favour of `show ... by ...`
- Inlined single-use haves and intermediate lemmas
- Converted tactic-block proofs to term-mode where short
- Replaced multi-step `calc` blocks with direct lemma applications
- Collapsed `refine ...; intro ...` into `refine ... fun x ↦ ...` and `intro x y; simp` into anonymous-constructor field syntax
- Reflowed definitions/statements to fit fewer lines within the 100-col limit

Part of this was done by hand, part with Claude's assistance; everything compiles as before.


kbuzzard:
Nice idea -- thanks!

Do you know for sure that your shorter proofs are always quicker? You can time them with `set_option trace.profiler true` (and I usually also use `set_option trace.profiler.useHeartbeats true` as well)

pepamontero:
@kbuzzard thank you! I know that some proofs are definitely a little bit slower, specially the simpa instead of rw+exact, but I have the impression these are preferred in terms of mantainability. But I'm not sure whether the `ring_nf` instead of `congr 1; ring` is actually better...

PR #1088, opened 2026-07-07
Quadratic twist to split multiplicative reduction

MichaelStollBayreuth:
Proves `WeierstrassCurve.exists_quadraticTwist_hasSplitMultiplicativeReduction`: over the fraction field `K` of a discrete valuation ring `R`, an elliptic curve with **nonsplit multiplicative reduction** acquires **split multiplicative reduction** after twisting by the unramified quadratic extension `L/K`. Proven in **both residue characteristics**, `sorry`-free (axioms: `propext, Classical.choice, Quot.sound`).

### FLT-specific files (`FLT/KnownIn1980s/EllipticCurves/QuadraticTwists/`)
- `QuadraticTwists.lean` — the quadratic twist: explicit Weierstrass model `quadraticTwistOf`/`quadraticTwistBy`/`quadraticTwist`, uniform in the characteristic; ellipticity, `j`-invariance, independence of choices, twisting as involution; the point-level isomorphism `quadraticTwistPointEquiv : Eᴸ(M) ≅ E(M)` with its Galois anti-equivariance `φ(σP) = χ(σ)·σ(φP)`; classification of the forms of `E` split by `L/K` (for `j ∉ {0, 1728}`).
- `SplitMultiplicativeReduction.lean` — the main theorem: the key residue identity `residue c₄ · residue(t²−4n) = −residue c₆` for the twisting datum, the split criteria applied to the twisted `R`-model (char ≠ 2 via square classes, char 2 via Artin–Schreier), and transfer to the minimal model.

### Mathlib-ready material (`FLT/Mathlib/`, one file per target Mathlib file)
Each file imports (essentially) only its Mathlib counterpart; where no existing file fits, the header declares a proposed new Mathlib file.
[...]

PR #1103, opened 2026-07-09
feat: tateEquiv_variableChange

WilliamCoram:
In this PR we prove `WeierstrassCurve.exists_variableChange_tateCurve`, which has now been moved to the `SplitMultiplicativeDescent.lean` file.

This PR needs lemmas from both #1088 and #1099, the statements of such lemmas are sorried in the SMD file, with pointers to the file and line they belong in their respective PR's. 
Note Claude earlier remarked it got half of the pointers wrong the first time... so who knows if that is really where they are. 

This should not be merged until after those PR's at which we can remove the duplicated sorries.

Co-authored by @smwyin (and Claude)


WilliamCoram:
This is being subsumed by our mega PR so will be closed

PR #1106, opened 2026-07-10
Tate equiv

SmwYin:
Proved one more sorry in `TateCurve.lean`.

Based on #1099

Co-authored by @WilliamCoram (and Claude + Codex)

PR #1109, opened 2026-07-10
feat: tateCurve file work

WilliamCoram:
This mega PR subsumes our now closed PR's #1103, #1099 and #1106; it completes our goal of the week, finishing the Tate Curve sorries up-to the Torsion and Weil pairing section.

This depends on #1088, with two sorried lemmas existing in SplitMultiplicativeReduction. These are either duplicates of lemmas existing there or deducible from lemmas. The dependent lemmas are labelled and point to the file and line of where they exist in that PR. Note the labels were depending on an old version, so will be updated.

We are both experimenting with cleaning the Tate curve work using Chris' skills.

We also introduce the MaybeMathlib file - this includes things we think should be in mathlib, but are not confident of putting them into the respective mathlib files until we do our clean.

Co-authored by @smwyin and ClaudeCode/Codex.


WilliamCoram:
Converting to draft until either myself or Sam works on bumping this post #1088 and attempting some cleaning.

PR #1098, opened 2026-07-09
feat(Mathlib/RepresentationTheory/Homological): define cup product for continuous cohomology

Whysoserioushah:

PR #1110, opened 2026-07-10
feat(KnownIn1980s): state Local Duality for finite discrete modules in continuous cohomology

Whysoserioushah:
depends on #1098


FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,118 @@
...
+
+public import FLT.Mathlib.RepresentationTheory.Homological.ContCohomology.CupProduct
+public import FLT.Mathlib.RepresentationTheory.Smooth.Basic
+public import Mathlib.NumberTheory.LocalField.Basic
+public import FLT.Deformations.RepresentationTheory.AbsoluteGaloisGroup
+
+/-!
+# The local Tate duality pairing
+
+Let `K` be a nonarchimedean local field of characteristic zero and let `M` be a finite discrete
```
kbuzzard, line 16: Why characteristic zero?

FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,118 @@
...
+variable (K M : Type u) [Field K] [CharZero K] [ValuativeRel K] [TopologicalSpace K]
+  [IsNonarchimedeanLocalField K] [TopologicalSpace M] [AddCommGroup M] [DiscreteTopology M]
+  [Finite M] (ρ : ContRepresentation ℤ (Field.absoluteGaloisGroup K) M)
+  [hρ : Representation.Smooth.IsSmooth ρ.toRepresentation]
+
+open TopRep
+
+/-- The group `(AlgebraicClosure K)ˣ`, written additively, carries the discrete topology. -/
+instance instTopologicalSpaceAdditiveUnitsAlgebraicClosure :
+    TopologicalSpace (Additive (AlgebraicClosure K)ˣ) := ⊥
```
kbuzzard: This is not a great idea. First, you're fixing an algebraic closure instead of using all of them. Second, it should be a separable closure. Third, this space already has a topology coming from the metric on K, and the topology it already has is the correct one; this is the incorrect one. What is going on here? 

I know Richard is insisting on putting topologies on everything but I really want to push back against this. K-bar-* already has a topology and if you're saying that it's the wrong topology then I want to argue that it's your theory which is wrong, not the topology. This is a really important point. I have discussed this with Richard several times and I have been wholly unconvinced by his answers that the theory you are developing is used anywhere at all in mathematics. This instance is no good, and if you need it then your theory is problematic. What we need for FLT is Galois cohomology, where the group is profinite and the module *has no topology* -- I don't mean "has the discrete topology", I mean *we do not ask for a topology so it can have any topology*, but stabilisers are open. I would like to see an example in mathematics where this hypothesis fails before I start allowing putting the discrete topology on objects which already have a topology which is not discrete; this is asking for trouble.

Whysoserioushah: Do we already have a topology on K_bar in FLT/Mathlib? Is the unit topology the right topology on K_bar^*? I could put this as an assumption if it's not there but then there would be more sorrys because I need continuity condition on pair₀ by my construction of cup product, and putting discrete topology made it easy, but if you're arguing that discrete topology is not the right topology then probably we need to pay the price by doing more hard work to prove the continuity.

FLT/KnownIn1980s/LCFT/Pairing.lean
Whysoserioushah: fixed

FLT/KnownIn1980s/LCFT/Pairing.lean
Whysoserioushah: should be fixed now

FLT/KnownIn1980s/LCFT/Pairing.lean
kbuzzard: In what sense?

FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,120 @@
...
+Let `K` be a nonarchimedean local field of characteristic zero and let `M` be a finite discrete
+module over the absolute Galois group `Gal(K̄/K)` of `K`. This file constructs the local Tate
+duality pairing on the continuous group cohomology of `Gal(K̄/K)`, as the cup product induced by
+the evaluation pairing of `M` with its dual `M^∨ = Hom(M, K̄ˣ)`.
+
+## Main definitions
+
+* `ContRepresentation.multiplicative K`: the continuous representation of `Gal(K̄/K)` on the
+  multiplicative group `K̄ˣ` of the algebraic closure, written additively and carrying the
+  discrete topology.
```
kbuzzard, line 25: I don't think you need this. Firstly I don't think you should be using Kbar, I think you should be using an arbitrary separable closure of K. Secondly, I don't think you should be putting topologies on things which might in future inherit different topologies.

FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,120 @@
...
+* `LocalGaloisRep.dual K M ρ`: the dual `M^∨ = Hom(M, K̄ˣ)` of a local Galois representation
```
kbuzzard, line 26: I don't think you need this. I think you should have `PerfectDuality M N` with `M` and `N` G-modules, which is defined to be a bilinear map b : M -> N -> Additive (KSep^*) (additive group homs) satisfying b(gm,gn)=gb(m,n). This is more general and will be more useful.

FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,120 @@
...
+  `ρ` on `M`, i.e. the internal hom of `M` into the multiplicative representation, with
+  `Gal(K̄/K)` acting by conjugation.
+* `pair₀ K M ρ`: the evaluation pairing `m ↦ (φ ↦ φ m)` of `M` with `M^∨`, as an intertwining
+  map from `ρ` into the internal hom of the dual and the multiplicative representation.
+* `pairing K M ρ i j hij`: the local Tate duality pairing
+  `Hⁱ(Gal(K̄/K), M) × Hʲ(Gal(K̄/K), M^∨) → H²(Gal(K̄/K), K̄ˣ)` for `i + j = 2`, defined as the
```
kbuzzard, line 32: I think we should also have the iso H^2(Gal(KSep/K),KSep^*)=Q/Z

FLT/KnownIn1980s/LCFT/Pairing.lean
```diff
@@ -0,0 +1,120 @@
...
+for `i + j = 2`: the cup product on continuous group cohomology induced by the evaluation
+pairing `pair₀`. -/
+noncomputable def pairing (i j : ℕ) (hij : 2 = i + j) :
+    continuousCohomology i (.of ρ) ⟶ TopModuleCat.linHom (k := ℤ)
+      (continuousCohomology j (LocalGaloisRep.dual K M ρ)) (continuousCohomology 2
+      (mkDiscrete (Rep.of (Representation.multiplicative K)))) :=
+  ContinuousCohomology.cup _ _ _ (pair₀ K M ρ)
+    (ContRepresentation.continuous_pair_of_discrete (pair₀ K M ρ)) i j 2 hij
+
+lemma pairing_isPerfect1 (i j : ℕ) (hij : 2 = i + j) :
```
kbuzzard, line 110: You are missing the hypothesis that the size of the G-module M is nonzero when considered as an element of K, otherwise H^i(Galois,M) might not be finite and the pairing might not be perfect.

PR #1113, opened 2026-07-10
Make CI enforce --iofail on lake test; fix FLTTest copyright headers

marcelolynch:
CI declares `test-args: --iofail` on the `lean-action` step, but lean-action silently drops the `test-args` input (leanprover/lean-action#163). So `lake test` has been running with no arguments, and warnings in test targets don't fail CI.

This PR:

1. Adds the `GITHUB_ENV` workaround (same as [cslib's](https://github.com/leanprover/cslib/commit/1d91c3f242d435f8d97b4511d254b35e82dea4c3)): a step exporting `TEST_ARGS='--iofail'`, which does reach `lake_test.sh`. The existing `test-args` input stays so behavior is unchanged when the action is fixed, at which point the workaround step can be deleted
2. Fixes the headers that would otherwise fail under the now-enforced flag

### DIFF
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -81,7 +81,7 @@ jobs:
       # Remove once leanprover/lean-action#153 is merged.
       - name: Set TEST_ARGS manually
         run: |
-          echo "TEST_ARGS='--iofail'" >> $GITHUB_ENV
+          echo "TEST_ARGS=--iofail" >> "$GITHUB_ENV"
         shell: bash
 
       - name: Build and lint project
--- a/.github/workflows/blueprint.yml
+++ b/.github/workflows/blueprint.yml
@@ -76,6 +76,14 @@ jobs:
         with:
           fetch-depth: 0 # Fetch all history for all branches and tags
 
+      # lean-action drops the test-args input (leanprover/lean-action#163), so
+      # export TEST_ARGS directly: lake_test.sh reads it from the environment.
+      # Remove once leanprover/lean-action#153 is merged.
+      - name: Set TEST_ARGS manually
+        run: |
+          echo "TEST_ARGS='--iofail'" >> $GITHUB_ENV
+        shell: bash
+
       - name: Build and lint project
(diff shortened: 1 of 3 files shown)


kbuzzard:
Claude says

***

  Two nits, neither blocking:

  The quotes in echo "TEST_ARGS='--iofail'" are load-bearing in a way that's a bit lucky. $GITHUB_ENV does not
  strip quotes, so the variable's literal value is '--iofail', quotes and all. It only works because the action's
  script uses eval, which re-parses the string and drops them. If lean-action ever changed that line to a plain
  lake test $TEST_ARGS, Lake would receive '--iofail' verbatim and reject it as an unknown option. echo 
  "TEST_ARGS=--iofail" >> "$GITHUB_ENV" is immune either way. (The quoted form is what cslib used, so I understand
  the choice — it's just strictly worse.) Same line: >> "$GITHUB_ENV" with quotes around the variable is the
  conventional safe form, though it makes no difference on a GitHub runner.

  One thing I'd genuinely like to see, if you want certainty rather than my reasoning: nothing in this PR
  demonstrates the flag is now enforced — a green CI run looks identical whether --iofail is active or still being
  dropped. Asking Marcelo to push a throwaway commit that adds e.g. #eval 1 to an FLTTest file, confirm CI goes
  red, then revert it, would prove the enforcement is live. I'm fairly confident it is (env var reaches the step,
  eval strips the quotes, Lake accepts the flag), but that's a chain of three inferences and the check is a
  two-minute job.

***

I am going to try and deal with these myself. Thanks a lot for this PR (which Claude is happy with).

kbuzzard:
@marcelolynch I added [this commit](https://github.com/ImperialCollegeLondon/FLT/pull/1113/changes/194e4d95863f79553548b1e03c92a49a9795139f) to your work and I'm now merging. Thanks. Let me know if you have any comments on the commit I added -- Claude did it and I know very little about CI.

marcelolynch:
@kbuzzard Looks good to me. I was just copying what cslib was doing so I don't think there was going to be an issue, but it still makes sense.

Issue #889, opened 2026-04-15
unfolding in typeclass isdefeq check for adicCompletion


kbuzzard:
In the proof of `IsDedekindDomain.HeightOneSpectrum.tensorAdicCompletionIntegersTo_isClopen_range` in `FLT/DedekindDomain/Completion/BaseChange.lean` and in several theorems in `FLT/DedekindDomain/FiniteAdeleRing/BaseChange.lean`, there are hundreds of occurences of the following issue.

It starts with what looks like a simple defeq check:
```
[isDefEq] [239348.000000] ✅️ Module K (adicCompletion K v) =?= Module K (adicCompletion K v) ▼
```

This gets unfolded until we see that actually there is a slight mismatch in some instances:
```
[] [228552.000000] ✅️ CommRing.toNonUnitalCommRing.toNonUnitalNonAssocCommRing.toAddMonoid =?= AddMonoidWithOne.unary.toAddMonoid ▼
```

These instances are equal under `with_reducible_and_instances` (indeed that's why the check is succeeding). The problem is that it's taking 200K mHeartbeats and it is happening again and again and again in this proof. The reason that it's so slow is because it expands to

```
[] [225038.000000] ✅️ UniformSpace.Completion.instAddMonoid =?= UniformSpace.Completion.instAddMonoid ▼
```

and then to

```
[] [224619.000000] ✅️ (WithVal.instCommRing (valuation K v)).toAddGroupWithOne.toAddGroup =?= 
(WithVal.instField (valuation K v)).toAddGroupWithOne.toAddGroup ▼
```

[...]

PR #1132, opened 2026-07-23
perf: remove the respectTransparency workarounds in the #889/#891 hotspot files

### LINKED ISSUE #891: slow typeclass inference for infinite place completions
```
[tryResolve] [10089531.000000] ✅️ IsTopologicalRing
                        ((vi : InfinitePlace K) →
                          D ⊗[K] vi.Completion) ≟ IsTopologicalRing ((i : InfinitePlace K) → D ⊗[K] i.Completion)
```

After the bump to v4.29, typeclass inference has got a lot slower on infinite place completions.

anagnorisis2peripeteia:
Removes 8 of the 9 `set_option backward.isDefEq.respectTransparency false` workarounds in the three hotspot files implicated by #889 and #891 (`DedekindDomain/Completion/BaseChange.lean`, `HaarMeasure/HaarChar/FiniteAdeleRing.lean`, `DivisionAlgebra/Finiteness.lean`), replacing each with a fix at the level the defeq check actually fails. Two supporting `FLT/Mathlib` files are touched (a new `rfl` simp lemma; a def-site attribute). Roughly 80 further sites elsewhere in the repo — including one in the supporting file `FLT/Mathlib/Algebra/Algebra/Pi.lean` itself (`Pi.algHomEquivOfIsDomain`), untouched here — are deliberately out of scope and could be attacked the same way as follow-up.

Closes #891. Towards #889.

## Why

These `set_option` lines are v4.29-migration escape hatches: they don't make the offending defeq checks cheap, they make unification stop respecting transparency for a whole declaration. #889 and #891 record the underlying disease (instance paths through type synonyms that stop being definitionally comparable at `.instances` transparency). This PR fixes the root causes for the hotspot files, in the style mathlib itself used for the same migration (mathlib4#35761: mark "pseudo-parent" definitions `@[implicit_reducible]`), so the workarounds can simply be deleted.

## What changed (per site)

| Site | Fix |
|---|---|
| `DedekindDomain/Completion/BaseChange.lean` (2 sites: `baseChangeRight_surjective`, `baseChangeContinuousAlgEquiv`) | culprit chain `semialgHomPi → Pi.semialgHom → RingHom.pi` opaque at instances transparency; `@[implicit_reducible]` on FLT-owned `semialgHomPi` and `Pi.semialgHom` at their definitions + `attribute [local implicit_reducible]` for mathlib-owned `RingHom.pi` |
| `HaarMeasure/HaarChar/FiniteAdeleRing.lean:120` (`CompactSpace` local instance) | `attribute [local implicit_reducible] AddSubgroup.pi AddSubmonoid.pi` |
| `HaarMeasure/HaarChar/FiniteAdeleRing.lean:359` (`localcomponent_matrix`) | new `@[simp] lemma AlgHom.rTensor_tmul … := rfl` (in `FLT/Mathlib/LinearAlgebra/TensorProduct/Algebra.lean`) + faithful `simp only` closes; no transparency attributes needed |
| `HaarMeasure/HaarChar/FiniteAdeleRing.lean:521` (`f_g_local_global`) | `attribute [local implicit_reducible] IsDedekindDomain.FiniteAdeleRing RestrictedProduct` + deterministic `simp only`/`Finset.sum_congr` finisher |
| `DivisionAlgebra/Finiteness.lean:533` (`tensorPi_equiv_piTensor_map_mul`) | `suffices` respelling so the statement stops mixing `K∞` with its Π-type unfolding |
| `DivisionAlgebra/Finiteness.lean:552` (`DinfTensorPiEquivPiTensorAux`) | term-level crossing (`rw [h₁, h₂]` + explicit `.trans`) instead of a transparency-dependent rewrite |
| `DivisionAlgebra/Finiteness.lean:931` (`toQuot_surjective`) | `attribute [local implicit_reducible] Subgroup.subgroupOf` |

**The one deliberate survivor in the hotspot files** — `Finiteness.lean` (`algebraMap_completion`): the failing pairs are `f i * g i =?= instCommRingInfiniteAdeleRing._aux_13 K f g i` (and the `_aux_1` addition twin) — machine-generated field auxiliaries of mathlib's `deriving CommRing` on `InfiniteAdeleRing`, with unstable names. Marking one aux just shifts the failure to the next. That needs a mathlib-side change (e.g. deriving via `inferInstanceAs (CommRing (Π v, v.Completion))`); the `set_option` stays with a comment recording exactly this, and I can file the mathlib issue as follow-up.

## Evidence

- **Each removed workaround is load-bearing on the base revision**: deleting any single one on `main` breaks its module's build (unsolved goals, rewrite-pattern misses, or type-mismatch elaboration errors, depending on the site). Verified site-by-site with a scripted audit (`lake build <module>` per site, files restored between sites).
- **HEAD is green with all 8 gone**: `lake build --iofail`, `lake test --iofail`, `lake lint`, `lake exe mk_all --check`, and the no-bare-`import Mathlib` style check all pass (`--iofail`: zero warnings).
- **#891's headline symptom is gone**: the ~10.09M-heartbeat `tryResolve` of `IsTopologicalRing ((vi : InfinitePlace K) → D ⊗[K] vi.Completion)` no longer occurs; the same resolution now costs ≈192K heartbeats (`trace.profiler` + `trace.profiler.useHeartbeats` on the formerly pathological declarations). Remaining synthesis in that region has no single pathological node.
- **Honest perf framing**: build wall/user times are parity within noise — this is defeq hygiene and tail-risk removal (correct elaboration without transparency escape hatches), not a speedup. The residual compile cost of these files is distributed TensorProduct-coercion defeq churn, which is a separate (partly mathlib-side) problem.
- Localization used `#defeq_abuse` (Mathlib.Tactic.DefEqAbuse); on the fixed declarations it now reports no abuse. Diagnostic scaffolding is fully removed from the diff.

## Test plan

```
lake exe mk_all --check
! (find FLT -name "*.lean" -type f -print0 | xargs -0 grep -E -n '^import Mathlib$')
env TEST_ARGS=--iofail lake build --iofail
lake test --iofail
lake lint
```

Review history: https://gist.github.com/anagnorisis2peripeteia/defa15335be43177b3805eb08f56c06d

---
Authored with Claude (Claude Code); every fix verified by the full local CI equivalent above.

### DIFF
--- a/FLT/DedekindDomain/Completion/BaseChange.lean
+++ b/FLT/DedekindDomain/Completion/BaseChange.lean
@@ -106,6 +106,14 @@ end assumptions
 
 variable [Algebra.IsIntegral A B] [IsFractionRing B L] [IsDedekindDomain B]
 
+/- The `K_v`-algebra structure on `∏_{w|v} L_w` is built through `adicCompletion.semialgHomPi`,
+`Pi.semialgHom` and `RingHom.pi`, so instance-implicit defeq checks (e.g. relating `ψ.toAlgebra`
+with the componentwise `Pi` instances) must be able to see through them at `instances`
+transparency (`backward.isDefEq.respectTransparency := true`, the default since Lean v4.29).
+`semialgHomPi` and FLT's own `Pi.semialgHom` are tagged `@[implicit_reducible]` at their
+definitions; mathlib-owned `RingHom.pi` gets a local attribute here. -/
+attribute [local implicit_reducible] RingHom.pi
+
 namespace IsDedekindDomain.HeightOneSpectrum
 
 variable (v : HeightOneSpectrum A) {A B}
@@ -251,6 +259,7 @@ namespace adicCompletion
 variable (B)
 
 /-- The canonical map `K_v → ∏_{w|v} L_w` extending K → L. -/
+@[implicit_reducible]
 noncomputable def semialgHomPi :
     v.adicCompletion K →ₛₐ[algebraMap K L] ∀ w : v.Extension B, w.1.adicCompletion L :=
   Pi.semialgHom _ _ fun i ↦ i.adicCompletionSemialgHom K L
@@ -291,7 +300,6 @@ noncomputable local instance :
     · apply ne_of_lt
       rwa [valuation_of_algebraMap, intValuation_lt_one_iff_mem]
 
-set_option backward.isDefEq.respectTransparency false in
 open scoped TensorProduct.RightActions in
 /-- The canonical map `L ⊗[K] K_v → ∏_{w|v} L_w` is surjective. -/
 lemma baseChangeRight_surjective [FiniteDimensional K L] :
@@ -779,7 +787,6 @@ noncomputable def baseChangeAlgEquiv :
     L ⊗[K] v.adicCompletion K ≃ₐ[L] Π w : v.Extension B, w.1.adicCompletion L :=
   AlgEquiv.ofBijective (baseChange K L B v) <| baseChange_bijective K L B v
 
-set_option backward.isDefEq.respectTransparency false in
 open scoped TensorProduct.RightActions in
 /-- The continuous L-algebra isomorphism `L ⊗[K] K_v ≅ ∏_{w|v} L_w`. -/
 noncomputable def baseChangeContinuousAlgEquiv :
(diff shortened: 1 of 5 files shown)


anagnorisis2peripeteia:
awaiting-review

PR #1134, opened 2026-07-24
fix: make the pre-push scripts work on a lakefile.toml checkout without leanblueprint

### LINKED ISSUE #1133: scripts/pre-push.sh checks for lakefile.lean, so the installed pre-push hook always fails
Two related problems with the pre-push tooling that CONTRIBUTING.md points contributors at:

1. `scripts/pre-push.sh` guards on `[ ! -f lakefile.lean ]`, but the project uses `lakefile.toml` — so after `scripts/install_pre-push.sh`, the hook aborts every push with "This doesn't appear to be the outermost 'FLT' directory". (`scripts/run_before_push.sh` already checks `lakefile.toml` correctly.)

2. Both scripts then hard-fail on `leanblueprint pdf` / `leanblueprint web` / `lake exe checkdecls blueprint/lean_decls` for anyone without `leanblueprint` installed — and the blueprint CI step is currently disabled anyway during the Verso migration (#1066), so the scripts fail even where CI would pass.

Proposed fix (happy to PR): correct the lakefile check, and skip the three blueprint steps with a notice when `leanblueprint` isn't on PATH, leaving behavior unchanged for blueprint-equipped users.

anagnorisis2peripeteia:
Fixes the two ways the pre-push tooling that CONTRIBUTING.md points contributors at fails on a standard checkout.

Closes #1133.

## What was broken

1. `scripts/pre-push.sh` guards on `[ ! -f lakefile.lean ]`, but the project uses `lakefile.toml` — so after running `scripts/install_pre-push.sh`, the git hook aborts **every push** with "This doesn't appear to be the outermost 'FLT' directory".
2. Both `pre-push.sh` and `run_before_push.sh` then hard-fail on `leanblueprint pdf` / `leanblueprint web` / `lake exe checkdecls` for anyone without `leanblueprint` installed — even though the blueprint CI step is currently disabled during the Verso migration (#1066), so the scripts fail where CI would pass.

## The fix

- `pre-push.sh`: check for `lakefile.toml` (matching `run_before_push.sh`, which already had it right).
- Both scripts: run the three blueprint steps only when `command -v leanblueprint` succeeds, otherwise print a skip notice. Behavior is unchanged for users who do have leanblueprint installed.

## Evidence (same scenario, base vs this branch, exit codes captured)

| Script | base (`8dd8088`) | this branch |
|---|---|---|
| `bash scripts/pre-push.sh` | exit 1 — aborts at the directory check | exit 0 — cache + `lake build FLT` green, blueprint steps skipped with a notice |
| `bash scripts/run_before_push.sh` | exit 1 — `leanblueprint: command not found` after a full successful build | exit 0 — same run, blueprint steps skipped with a notice |

## Test plan

```
bash scripts/pre-push.sh          # exit 0 on a lakefile.toml checkout without leanblueprint
bash scripts/run_before_push.sh   # exit 0, blueprint steps skipped with a notice
env TEST_ARGS=--iofail lake build --iofail && lake test --iofail && lake lint && lake exe mk_all --check
```

Review history: https://gist.github.com/041c358b97ca07ff31ef49a26988b428

---
Authored with Claude (Claude Code).

### DIFF
--- a/scripts/pre-push.sh
+++ b/scripts/pre-push.sh
@@ -13,7 +13,7 @@
 
 # Ensure the script is in the outermost 'FLT' folder
 echo "Checking if you are in the correct directory..."
-if [ ! -f lakefile.lean ]; then
+if [ ! -f lakefile.toml ]; then
   echo "❌ Error: This doesn't appear to be the outermost 'FLT' directory.
         Please run this script from the correct folder."
   exit 1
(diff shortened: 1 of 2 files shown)


anagnorisis2peripeteia:
awaiting-review

PR #1172, opened 2026-08-14
chore: bump mathlib to 9058eaf, fix breaking changes

kbuzzard:
Bumps mathlib from bc06ce9 to 9058eaf, 51 commits on.

* mathlib#41291 (`refactor(RingTheory/IntegralClosure/IsIntegral/Basic): use FaithfulSMul in
  isIntegral_algebraMap_iff`) replaced the explicit `Function.Injective (algebraMap A B)` argument
  of `isIntegral_algebraMap_iff` with a `[FaithfulSMul A B]` instance. We update FLT accordingly.

* mathlib#41415 (`feat(FieldTheory/AbsoluteGaloisGroup): functoriality of absoluteGaloisGroup`)
  upstreamed FLT's `Field.absoluteGaloisGroup.map`, so FLT's own copy (together with the `mapAux`
  it was built from) is deleted and `lift_map` is reproved against Mathlib's definition. Mathlib
  states `map` for arbitrary base fields rather than number fields, so the downstream lemmas that
  were silently picking up `[NumberField K]` through it now have to `omit` it.


kbuzzard:
<sub>Opened automatically by the nightly bump job on Kevin's machine (`~/.local/bin/flt-nightly-bump.sh`). Full log: `/home/buzzard/.local/state/flt-nightly-bump/run-20260814-040001.log`.</sub>

PR #1115, opened 2026-07-11
feat: prove injective_hRat and injective_zHat in HurwitzRatHat

theTyster:
Fills the two sorries marked `-- flatness` in `FLT/Data/HurwitzRatHat.lean`.

The missing piece was a `Module.Flat ℤ 𝓞` instance for the Hurwitz order. This PR:

- adds an `IsAddTorsionFree 𝓞` instance in `FLT/Data/Hurwitz.lean`, transported along the injective `toQuaternion` via a new bundled `toQuaternionAddHom : 𝓞 →+ ℍ` (ℍ is torsion-free as an ℝ-module, and torsion-free over the Dedekind domain ℤ gives flatness via existing mathlib instances);
- adds `Module.Flat ℤ` instances for `𝓞^` and `D` (tensor products of flats);
- proves `injective_hRat` (`lTensor` of the injective `ℤ → ℤ̂` into the flat `D`) and `injective_zHat` (an `injective_includeRight` helper mirroring `QHat.injective_zHat`, composed with the injective `assoc.symm`).

The two remaining sorries (`canonicalForm`, `completed_units`) are genuinely different (denominator-clearing / units machinery) and are left for follow-ups.

Verification: `lake build FLT` green on top of current `main` (post-#1100 mathlib bump); the four new declarations depend only on `propext`, `Classical.choice`, `Quot.sound`.

PR #1183, opened 2026-08-26
feat: prove canonicalForm in HurwitzRatHat

theTyster:
This fills the `canonicalForm` sorry in `FLT/Data/HurwitzRatHat.lean`. Every element of `D^` can be written as `j₁ (1/N) * j₂ z'` for some `N : ℕ+` and `z' : 𝓞^`.

It's basically the quaternionic version of `QHat.canonicalForm`, and it follows the same shape. Induct with `TensorProduct.induction_on`, clear denominators on the pure tensors by taking `N = q.den` and `z' = q.num • o`, then put sums over a common denominator.

Two things are different, because `D^` is `(ℚ ⊗ 𝓞) ⊗ ℤ̂` instead of `ℚ ⊗ ℤ̂`. The pure tensor case needs a second induction, over `D = ℚ ⊗ 𝓞`. And the common denominator step is factored out into `canonicalForm_add`, since it's needed in both `add` branches now. `j₁_smul_aux` is the rescaling identity it leans on.

Both helpers are private, so the public API doesn't change. The diff takes out exactly one line, the `sorry`, and the statement is as you wrote it.

On length: I know this is longer than `QHat.canonicalForm`, so I wanted to flag it myself. `canonicalForm` itself is about the same, 23 lines against 22. The extra is the two helpers. Most of what's left is that `D`, `D^` and `𝓞^` are `def`s and not `abbrev`s, so `simp` and `rw` won't see through `j₁` and `j₂`. The `change` in the pure tensor case and the term mode `congrArg`/`.trans` step in `j₁_smul_aux` are both there to get around that. It isn't for want of golfing. I'm happy to restructure it if you'd rather it read a different way. What do you think?

This one is independent of #1115 and #1178, which touch the two `-- flatness` sorries further up the same file. It only touches `canonicalForm`. The last sorry in the file, `completed_units`, is the units machinery, and I left that for a follow-up.

Verification: `lake build FLT` is green on current `main` (mathlib `f006b2e`, `v4.34.0-rc2`), 9052 jobs. `canonicalForm` depends only on `propext`, `Classical.choice` and `Quot.sound`.

### DIFF
--- a/FLT/Data/HurwitzRatHat.lean
+++ b/FLT/Data/HurwitzRatHat.lean
@@ -89,9 +89,55 @@ noncomputable abbrev j₂ : 𝓞^ →ₐ[ℤ] D^ :=
 lemma injective_zHat :
     Function.Injective j₂ := sorry -- flatness
 
+/-- Scaling the `𝓞^`-part by `M` cancels the factor `M` in the denominator `N * M`. -/
+private lemma j₁_smul_aux (N M : ℕ+) (w : 𝓞^) :
+    j₁ ((((N * M : ℕ+) : ℚ)⁻¹) ⊗ₜ 1 : D) * j₂ ((M : ℤ) • w)
+      = j₁ (((N : ℚ)⁻¹) ⊗ₜ 1 : D) * j₂ w := by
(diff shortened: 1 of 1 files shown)


theTyster:
awaiting-review

PR #1188, opened 2026-09-03
chore: bump mathlib to c4a007f, fix breaking changes

kbuzzard:
Bumps mathlib from 1055fda to c4a007f, 140 commits on.

* mathlib#43061 (`feat(GroupTheory): results on normal subgroups of nilpotent groups`) added
  `Subgroup.normal_of_le_center` upstream, taking the subgroup implicitly rather than explicitly.
  We delete FLT's own copy and update its call sites.

* mathlib#42787 (`chore(RingTheory/Ideal/Norm): weaken Ideal.absNorm to infinite Dedekind domains`)
  swapped the `[Module.Free ℤ S]` hypothesis of `Ideal.absNorm` for `[Infinite S]`. FLT's
  `Rat.HeightOneSpectrum` material now supplies that instance from the isomorphism `R ≃+* ℤ`; it is
  `local` because its discrimination key would otherwise be the bare `Infinite _`.

* mathlib#43078 (`refactor: Quaternion as abbrev`) lets the `QuaternionAlgebra` simp set fire
  through `Quaternion`, so two `simp`s in `FLT/Data/Hurwitz.lean` now close their goals outright and
  the trailing `rfl`s have become dead tactics. We drop them.

* The bump also flips which of two definitionally equal `MulOneClass GL₂(𝔸ᶠ[F])` instance paths Lean
  picks in `LevelStruct.inf_U_eq_iff`, so `convert` is left with a residual instance-equality goal;
  `rfl` closes it. We could not attribute this to a specific upstream PR — locally reverting the
  instance-priority changes of mathlib#40156 does not restore the old behaviour.


kbuzzard:
<sub>Opened automatically by the nightly bump job on Kevin's machine (`~/.local/bin/flt-nightly-bump.sh`). Full log: `/home/buzzard/.local/state/flt-nightly-bump/run-20260903-040001.log`.</sub>

PR #1190, opened 2026-09-05
chore: bump mathlib to 7974e75, fix breaking changes

kbuzzard:
Bumps mathlib from c4a007f to 7974e75, 72 commits on.

* mathlib#43382 (`chore(GroupTheory): make arguments implicit in two iff lemmas`) made the
  subgroup and element arguments of `DoubleCoset.eq` implicit. We update FLT's two call sites
  that passed them explicitly; the sites that already used `..` are unaffected.

* mathlib#36275 (`feat: norm for the finite adele ring of a number field`) split a new file
  `Mathlib/NumberTheory/NumberField/FiniteAdeleRing.lean` out, moving the `𝔸ᶠ[K]` notation
  there from `Mathlib/NumberTheory/NumberField/AdeleRing.lean`, and dropped
  `Mathlib.LinearAlgebra.FreeModule.IdealQuotient` from the imports of
  `Mathlib/NumberTheory/NumberField/Completion/FinitePlace.lean`. FLT was relying on both of
  those transitively, so we add the two imports it now needs directly.


kbuzzard:
<sub>Opened automatically by the nightly bump job on Kevin's machine (`~/.local/bin/flt-nightly-bump.sh`). Full log: `/home/buzzard/.local/state/flt-nightly-bump/run-20260905-040001.log`.</sub>

PR #1192, opened 2026-09-09
chore: bump mathlib to d9ed2b0, fix breaking changes

kbuzzard:
Bumps mathlib from 71a8058 to d9ed2b0, 90 commits on.

* mathlib#43481 (`chore(LinearAlgebra/TensorProduct): remove `zero` step from
  `TensorProduct.induction_on``) renamed `TensorProduct.induction_on` to
  `TensorProduct.inductionOn` and dropped its redundant `zero` step, both from the term-level
  lemma and from the `induction_eliminator` used by bare `induction x with`. We update FLT
  accordingly, deleting the now-nonexistent `zero` cases.

* mathlib#43414 (`chore(RingTheory/Ideal/GoingUp): use `Ideal.under``) restated the going-up
  lemmas in terms of `Ideal.under`. We update FLT to the new names
  (`Ideal.isMaximal_under_of_isIntegral_of_isMaximal`,
  `Ideal.isMaximal_of_isIntegral_of_isMaximal_under`, `Ideal.eq_bot_of_under_eq_bot`) and drop
  the `rw [← Ideal.under]` steps that are now redundant. Note that the two `isMaximal` names
  were *reused* upstream for the unbundled `RingHom` variants rather than deprecated, so the
  old call sites failed with type errors rather than deprecation warnings.

* mathlib#42804 (`refactor: use `Set.Pairwise s Commute` for commutativity of subobject
  closures`) changed `Algebra.isMulCommutative_adjoin` to take `s.Pairwise Commute`. We update
  the `HeckeAlgebra` instance to discard the extra `x ≠ y` hypothesis and to produce `Commute`
  rather than an equation.

* mathlib#27135 (`feat: `gconvert` tactic`) deprecates `peel` in favour of `gconvert`. We
  update FLT's two uses.

* mathlib#43428 (`chore: make the `coe` for `Valuation.Completion` simp normal`) changed the
  `Coe R v.Completion` instance to unfold to `WithVal.toVal`, so `WithVal.equiv_symm_apply` no
  longer fires. We drop it from the affected `simp only` and `rw` calls.


kbuzzard:
<sub>Opened automatically by the nightly bump job on Kevin's machine (`~/.local/bin/flt-nightly-bump.sh`). Full log: `/home/buzzard/.local/state/flt-nightly-bump/run-20260909-040001.log`.</sub>

